PR #130, opened 2022-02-07
Ref lookups

nataliemisasi:
New references/lookups article

Not sure what I did - but Anais' newsletter updates are in here. I must have reverted her changes as well 👎  -- maybe not, I saw them but not anymore


help/references-lookups.md
```diff
@@ -186,13 +190,13 @@ Every row has a numeric id (available as `$id` in formulas) that is unique withi
 <span class="screenshot-large">*![row-id](images/references-lookups/row-id.png)*</span>
 {: .screenshot-half }
 
-Let's take a look at the Registrants column of the Events table. 
+Let's take a look at the Registrants column of the [Events](https://public.getgrist.com/6kTypo2FtSsf/Event-Sponsors-Attendees-References-and-Lookups/p/3) table. 
 
 <span class="screenshot-large">*![lookup-records-id](images/references-lookups/lookup-records-id.png)*</span>
 
-The formula used here is `Attendees.lookupRecords(Event=$id)`. We use the ID to find a match rather than a column because there is an existing reference connecting the two tables.
+The formula used here is `All_Registrations.lookupRecords(Event=$id)`. We use the ID to find a match rather than a column because there is an existing reference connecting the two tables.
```
anaisconce, line 197: Suggested edit (technically `$id` is a field like any other field).

> The formula used here is `All_Registrations.lookupRecords(Event=$id)`. We use the ID to find a match because in the All  Registrations table, the Event column is a *reference column* which means its *value* is a record's id.

help/references-lookups.md
```diff
@@ -186,13 +190,13 @@ Every row has a numeric id (available as `$id` in formulas) that is unique withi
...
 
-Since Attendees.Event is a reference column pointing to an Event record in the Events table, then it actually stores the unique ID of this Event. So in a formula for Registrants, using the lookup we see here, we will find all Registrant records tied to the current Event. 
+Since All_Registrations.Event is a reference column pointing to an Event record in the Events table, then it actually stores the unique ID of this Event. So in a formula for Registrants, using the lookup we see here, we will find all Registrant records tied to the selected Event.
```
anaisconce, line 199: Editing so it flows with my previous edit suggestions. 

> Because All_Registrations.Event is a reference column pointing to an Event record in the Events table, we can match the id stored in the reference column to the ids of records in the Events table. That's why the argument in the formula is `Event=$id`.

anaisconce:
This is looking great!! A few minor comments:

1. In `Reference Lists and Dot Notation`, maybe we can come up with a more realistic example. What if it just pulled registrant's emails? 

2. After explaining how to use sum and len on record sets, maybe useful to mention you can also use other functions like average, min, max, etc, and link to the function reference, `functions.md`

anaisconce (approved):

PR #138, opened 2022-03-16
document new add-or-update endpoint

paulfitz:
Screenshot from 2022-03-16 18-13-37


api/grist.yml
```diff
@@ -666,6 +666,64 @@ paths:
...
+            description: "Set to true to allow `require` in the body to be empty."
+      requestBody:
+        description: >
+          The records to add or update. Instead of an id,
+          a `require` object is provided, with the same structure as
+          fields. If no query parameter options are set, then the operation
+          is as follows. First, we check if a record exists matching
+          the values specified for columns in `require`. If so, we update
+          it by setting the values specified for columns in `fields`. If
+          not, we create a new record with the values specified in `fields`.
```
alexmojaki, line 717:
```suggestion
          not, we create a new record with a combination of the values
          in `require` and `fields`, with `fields` taking priority if the same column
          is specified in both.
```

api/grist.yml
```diff
@@ -666,6 +666,64 @@ paths:
...
+        - in: query
+          name: allow_empty_require
+          schema:
+            type: boolean
+            description: "Set to true to allow `require` in the body to be empty."
+      requestBody:
+        description: >
+          The records to add or update. Instead of an id,
+          a `require` object is provided, with the same structure as
+          fields. If no query parameter options are set, then the operation
```
alexmojaki, line 713:
```suggestion
          `fields`. If no query parameter options are set, then the operation
```

api/grist.yml
```diff
@@ -965,6 +1023,43 @@ components:
...
+      required:
+        - records
+      properties:
+        records:
+          type: array
+          items:
+            type: object
+            required:
+              - require
+              - fields
```
alexmojaki, line 1037:
```suggestion
```

Only `require` is required.

alexmojaki (changes requested):

api/grist.yml
```diff
@@ -666,6 +666,64 @@ paths:
...
+        - in: query
+          name: onmany
+          schema:
+            type: string
+            enum:
+              - first
+              - none
+              - all
+            description: >
+              When updating, which matches to update:
```
alexmojaki, line 685:
```suggestion
              Which records to update if multiple records are found to match `require`.
```

api/grist.yml
```diff
@@ -666,6 +666,64 @@ paths:
...
+                * `first` - at most one match
```
alexmojaki, line 686:
```suggestion
                * `first` - the first matching record (default)
```

api/grist.yml
```diff
@@ -666,6 +666,64 @@ paths:
...
+        - in: query
+          name: noupdate
+          schema:
+            type: boolean
+            description: "Set to true to prohibit updating records."
+        - in: query
+          name: noparse
+          schema:
+            type: boolean
+            description: "Set to true to prohibit converting types."
```
alexmojaki, line 703:
```suggestion
            description: "Set to true to avoid parsing strings according to the column type."
```

Several endpoints actually support this, but it's not mentioned in the others.

api/grist.yml
```diff
@@ -666,6 +666,64 @@ paths:
...
+        - in: query
+          name: allow_empty_require
+          schema:
+            type: boolean
+            description: "Set to true to allow `require` in the body to be empty."
```
alexmojaki, line 708:
```suggestion
            description: "Set to true to allow `require` in the body to be empty, which will match and update all records in the table."
```

api/grist.yml
paulfitz: Updated text (thanks) and documented noparse for /records POST+PATCH and /data POST+PATCH

paulfitz:
Thanks for all the clarifications @alexmojaki. Integrated them.

alexmojaki (approved):
Thanks!

PR #140, opened 2022-03-29
Formula Cheat Sheet

nataliemisasi:


dsagal:
I pushed up a change to update the styles of the formula cheat sheet; see https://deploy-preview-140--grist-help-preview.netlify.app/formula_cheat_sheet/. Let me know what you think! (I haven't had time to read through the content yet.)

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+
+<span class="screenshot-large">*![comparing-values-stock-alert](images/formula-cheat-sheet/comparing-values-stock-alert.png)*</span>
+
+The formula used here is:
+```
+if $In_Stock + $QTY_on_Order > 5:
+  return "In Stock"
+if $In_Stock + $QTY_on_Order > 0:
+  return "Low Stock"
+if $In_Stock + $QTY_on_Order <= 0:
```
dsagal, line 137: This is superfluous now, because the only way to get here is if this condition is already true. So simply an unindented `return "OUT OF STOCK"` would suffice. Maybe OK to keep but mention this fact in the text? Or change the formula so that we recommend the cleanest approach.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+
+The formula used in the Orphaned? column of the Orphaned Pages table is:
+```
+len(Links.lookupRecords(To=$id))<1
+```
+We use the [lookupRecords](https://support.getgrist.com/functions/#lookuprecords) function to find all records in the Links table where the link in the To column matches the link listed in the Slug column of this row.
+
+We use [len()](https://support.getgrist.com/functions/#len) to count the number of records found. If it's less than 1, the formula is evaluated to be true and the checkbox will be checked. If it's equal to or greater than 1, the formula is evaluated to be false.
+
+The No Internal Links? column of the [Pages Without Links to Other Pages](https://templates.getgrist.com/j9ZH7rPGafbH/Internal-Links-Tracker-for-SEO/p/5#a1.s10.r2.c33) table is another example of comparing values.
```
dsagal, line 172: This and the two examples above all take `len(X.lookupRecords(...))` and check if the result is 0, either `>0` or `<1`. Another way is to use `X.lookupRecords(...)` directly in comparisons, since it's truthy when non-empty, and falsy when empty. It can also be converted to a boolean value using `bool(X.lookupRecords(...))`. I think having one example would be good (e.g. `len(...) > 0`), but 3 examples make it sounds like it's a particularly recommended approach. An example where the comparison is unavoidable would be better, e.g. here https://community.getgrist.com/t/ensure-unique-values-or-detect-duplicates/76 (where it's `len(...) > 1`, i.e. 2 or more).

dsagal: Also, having a "Troubleshooting" section would be useful. It's an area where I run into trouble myself, whenever the two things being compared aren't both numbers.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+</details>
+</section>
+
+
+<span></span><section class="cheat-sheet">
+#### Converting from String to Float
+
+**[String](https://www.w3schools.com/python/python_strings.asp)**: A sequence of characters that once defined, cannot be changed. See [Python str() Function](https://www.w3schools.com/python/ref_func_str.asp) for converting a specified value to a string.
```
dsagal, line 180: I would describe a string more like https://futurecoder.io/course/#IntroducingStrings does: as snippets of text, or a sequence of characters. In code, strings are quoted e.g. `'Hello!'` or `"-12"` (the three characters in quotes, as opposed to a negative number).

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+
+**[Integer](https://www.w3schools.com/python/python_numbers.asp)**: A whole number, without decimals. See [Python int() Function](https://www.w3schools.com/python/ref_func_int.asp) for converting a specified value to an integer number.
```
dsagal, line 182: I'd move float up, since it is what's relevant to this recipe. I wonder if it's better to omit mention of integer. Do you think it helps, or adds confusion?

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+
+**[Float](https://www.w3schools.com/python/gloss_python_float.asp)**: Real numbers that can store decimal values. Also called floating point number. See [Python float() Function](https://www.w3schools.com/python/ref_func_float.asp) for converting a specified value into a floating point number.
+
+<span></span><details><summary>
+#### Example converting a string to a float
+</summary>
+
+**[Mixing Products](https://public.getgrist.com/v4vj2PDZS4jf/Community-665/m/fork)**
```
dsagal, line 190: With this example, it's hard to say what common problem it will help to solve. It's complicated and seems to solve some problems that are caused by having a column with numbers incorrectly set to Text type. 

That might indeed be a common problem, and maybe good to have a recipe, e.g. for dealing with numerical values in a Text column that contains a legitimate mix of numerical and text values. For the other direction (including numbers into a formatted string) this recipe can just link to the formatting recipe below.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+Specify the number of decimal places to give in a result using the [ROUND()](https://support.getgrist.com/functions/#round) function. If Average Temperature = `46.5`, `round($Average_Temperature)` would return `47`
+
+<span></span><details><summary>
+#### Example of rounding numbers
+</summary>
+**[Mixing Products](https://public.getgrist.com/v4vj2PDZS4jf/Community-665/m/fork)**
+
+<span class="screenshot-large">*![round](images/formula-cheat-sheet/round.png)*</span>
+
+The formula used in the Rounded Value column is:
```
dsagal, line 229: Here too, we could find an example that's easy to understand. Maybe https://templates.getgrist.com/5pHLanQNThxk/Payroll/p/2, "Payment" column, which uses Round with an argument because dollar amounts on paychecks should not include any sub-penny amounts.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+
+Finally, we use the [join()](https://www.w3schools.com/python/ref_string_join.asp) method to combine our two lists. `' '` is our starting (empty) string. We use Python's [format method](https://www.w3schools.com/python/ref_string_format.asp) to format our string. `{}` is a placeholder for each variable listed in `.format()`. Last, we use Python's [zip() function](https://www.w3schools.com/python/ref_func_zip.asp) to pair the first values from each list together and then pair the second values in each list together. `l` is assigned as our `first` list and `round_x` is assigned as our `second` list. `l = [Prod A, Prod B]` and `round_x = [2.36, 12.58]`. Zipping our lists into `'{} {}'.format(first, second)` gives us `Prod A 2.36` in our first iteration and `Prod B 12.58` in our second iteration. Our final return value is `Prod A 2.36 Prod B 12.58`.
+</details>
+</section>
+
+
+<span></span><section class="cheat-sheet">
+#### Specify Minimum Number of Digits
```
dsagal, line 256: Maybe call this "Formatting numbers with leading zeroes"? (Could also link from here to the other formatting recipe(s))

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+
+Allows you to specify the minimum number of digits returned in a numerical column by adding leading zeros. If x = 5, `str(x).zfill(3)` or `str('{:0>3}').format(x)` would return `005`.
```
dsagal, line 258: `str('{:0>3}').format(x)` should just be `'{:0>3}'.format(x)`

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+#### Example of Specifying Minimum Number of Digits
+</summary>
+
+**Community Example: [Using Row ID](https://public.getgrist.com/p4zvsX9asVCc/2179-Using-id/p/1)**
+
+<span class="screenshot-large">*![min-digits](images/formula-cheat-sheet/min-digits.png)*</span>
+
+The formula used in the 5-digit ID column of the ID Examples table is:
+```
+"TCH" + str('{:0>5}'.format($id))
```
dsagal, line 270: A better formula would be `'TCH {:0>5}'.format($id)`.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,913 @@
...
+`str($id)` converts the row ID to a string. `.zfill(x)` returns a copy of the string with leading zeros to make a string of length `x`. In our example, it adds leading zeros to make the string 5 characters in length. Again, if the `$id` is longer than 5 digits, the formula returns the string unmodified. 
+
+</details>
+
+<span></span><details><summary>
+#### Troubleshooting Errors
+</summary>
+`#TypeError`: can only concatenate str (not "int") to str
+
+If you mean to use the row ID as a character rather than a numerical value, be sure to convert it to string using `str()`.
```
dsagal, line 292: Maybe: If you mean to combine a string and a numerical value, be sure to convert the number to a string using `str()` (or use string formatting [link to recipe])

dsagal (commented):
I went over the Numerical portion of the formula cheat sheet. Good stuff there, I think it will be very helpful to people.

nataliemisasi:
> I pushed up a change to update the styles of the formula cheat sheet; see https://deploy-preview-140--grist-help-preview.netlify.app/formula_cheat_sheet/. Let me know what you think! (I haven't had time to read through the content yet.)

I like it! Much easier to see each subsection (and also the code is SO much cleaner).

help/formula_cheat_sheet.md
nataliemisasi: I'll move Float up. I thought having integer might be helpful since it's so closely related and our errors include int and float. At some point, I can create an integer example for this section and adjust the title. I made a note on my quip doc to add it later.

help/formula_cheat_sheet.md
nataliemisasi: > some problems that are caused by having a column with numbers incorrectly set to Text type.

I addressed this in the Troubleshooting section - let me know what you think

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+
+You'll notice that `Butter` is a duplicate within the Kitchen Inventory column but the cell is red, not orange. Conditional formatting runs top to bottom. First, we check for duplicates within the column and because `butter` has a duplicate, it becomes orange. Then, we check for duplicates across both columns and because `butter` is also a duplicate here, it then becomes red.
+
+#### Find Duplicates Across Multiple Columns Using a Helper Column
+<span class="screenshot-large">*![duplicates-multiple-columns](images/formula-cheat-sheet/duplicates-multiple-columns.png)*</span>
+
+The formula used in the Duplicates Across Columns? column of the Duplicates Across Multiple Columns table is:
+
+```
+"True" if len(Duplicates_Across_Multiple_Columns.lookupRecords(Grocery_List=$Kitchen_Inventory)) or len(Duplicates_Across_Multiple_Columns.lookupRecords(Kitchen_Inventory=$Grocery_List))>0 else ""
```
dsagal, line 716: No need for `"True" if` and `else ""`.

Also the long scrolling single line brings the worst memories of Excel: you can break it up across lines by enclosing in parentheses. E.g.
```
(len(Table1.lookupRecords(...)) > 0 or
 len(Table1.lookupRecords(...)) > 0 )
```
Making the table name shorter would make it a little easier to  read, as well as spaces around `>`, and using `> 0` consistently in both lookups.

To be honest, I would omit all these solutions for duplicates across multiple columns from the cheat sheet. They are very long and complicated, and I think it's only come up very rarely. For the purpose of the cheat sheet, it seems better to focus on more common scenarios.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+
+<span class="screenshot-large">*![id-reference-class-code](images/formula-cheat-sheet/id-reference-class-code.png)*</span>
+
+A reference column points to the <em>entire</em> record, not just the value you see here in the Class column. Using the configuration panel on the right hand side of the screen, you can pick any column from the originating table to show. For this example, the Class column shows the value from the Class Code column of the Classes table but it points to the entire record where the class code is `2018F-Stars`. 
+
+<span class="screenshot-large">*![id-class-id](images/formula-cheat-sheet/id-class-id.png)*</span>
+
+As you can see in this screenshot, the Row ID for this particular record is `1` and because we are calculating the Count for the row with `Row ID = 1`, it will count all records in the Enrollment table where Class shows `2018F-Stars` and Status is `Confirmed`.
```
dsagal, line 853: The `1` in the description here doesn't seem to match what's in the screenshot.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+
+<span class="screenshot-large">*![removing-duplicates](images/formula-cheat-sheet/removing-duplicates.png)*</span>
+
+The formula in the All Divisions column of the Abroad Trips table is:
+```
+confirmed_div = $Attending_Confirmed.Role_Division
+pending_div = $Attending_Pending.Role_Division
+full_list = confirmed_div + pending_div
+result = []
+[result.append(div) for div in full_list if div not in result]
```
dsagal, line 877: I suggest changing this line from a list comprehension to a loop:
```
confirmed_div = $Attending_Confirmed.Role_Division.Division
pending_div = $Attending_Pending.Role_Division.Division
full_list = confirmed_div + pending_div
result = []
for div in full_list:
  if div not in result:
    result.append(div)
return sorted(result)
```
But the `set()` approach is better here, and more efficient. (For very large lists, the set() approach would be way faster than the append() one)

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+
+As you can see in this screenshot, the Row ID for this particular record is `1` and because we are calculating the Count for the row with `Row ID = 1`, it will count all records in the Enrollment table where Class shows `2018F-Stars` and Status is `Confirmed`. 
+</details>
+</section>
+
+
+<span></span><section class="cheat-sheet">
+#### Removing Duplicates From a List
+
+You can remove duplicates from a list with help from Python's [append()](https://docs.python.org/3/tutorial/datastructures.html) method.
```
dsagal, line 861: A great way to remove duplicates is with the Python `set()` function. E.g. I added an example in column A of this copy here: https://public.getgrist.com/3pZUMdP2bJx6~d2zkxNHqAuEH9b8yC4nwub~5/1957/p/3

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+</details>
+</section>
+
+
+<span></span><section class="cheat-sheet">
+#### Setting Default Values for New Records
+
+You can set default values for when a new record is created and save yourself the trouble of having to fill in the same fields with the same values time after time.
+
+**Read about it in the Community:** [Default values on the widget](https://community.getgrist.com/t/default-values-on-the-widget/689/4)
```
dsagal, line 905: Another place where I think opening link in a new window is better.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+Another possibility is that this was entered in as a Formula column rather than a trigger formula column. Convert it to a trigger formula and this should resolve the problem.
+
+</details>
+</section>
+
+
+<span></span><section class="cheat-sheet">
+#### Filtering Data within a Specified Amount of Time
+
+Using the [`EDATE()`](https://support.getgrist.com/functions/#edate) function and [comparision operators](#comparing-values), you can determine if a date falls within a specific range then apply a filter.
```
dsagal, line 965: What do you think of recommending the `DATEADD()` function instead? `EDATE` is an Excel function that some people might already know, but `DATEADD` is Grist-specific and more general -- e.g. you could add/subtract days or years or weeks, or a combination.

help/formula_cheat_sheet.md
```diff
@@ -0,0 +1,1005 @@
...
+<span></span><section class="cheat-sheet">
+#### Setting Default Values for New Records
+
+You can set default values for when a new record is created and save yourself the trouble of having to fill in the same fields with the same values time after time.
+
+**Read about it in the Community:** [Default values on the widget](https://community.getgrist.com/t/default-values-on-the-widget/689/4)
+
+</section>
+
+## Working with dates and times
```
dsagal, line 909: I like this whole section, the examples here are easy to understand.

dsagal (commented):
I am really sorry it took me so long to get to review https://github.com/gristlabs/grist-help/pull/140. I went through it all again, this time to the end.

It's a ton of work, thank you! If anything, I'd nudge towards shortening, and preferring bite-sized simpler recipes that are less intimidating. Sorry about the too many comments...

help/formula_cheat_sheet.md
```diff
@@ -109,16 +109,96 @@ SUM(Incoming_Order_Line_Items.lookupRecords(SKU=$id).Received_Qty)
...
+```
+If the current date is greater than the date given in the Due Date column **and** the value in the Status column is not equal to `Completed`, the formula returns `True`. If either of these statements is false, the formula returns `False`.
+
+**[Restaurant Custom Orders](https://templates.getgrist.com/e4gEm7dt4cgB/Restaurant-Custom-Orders) template**
+
+<span class="screenshot-large">*![inequality-restaurant-bom](images/formula-cheat-sheet/inequality-restaurant-bom.png)*</span>
+
+The trigger formula used in the BOM # column of the Bill of Materials table is:
+```
+MAX(o.BOM_ for o in Bill_Of_Materials.all if o.id != $id) + 1
```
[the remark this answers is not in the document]
nataliemisasi: I understand and agree. Moved to "Using a Record’s Unique Identifier in Formulas" section.

dsagal (approved):
Looks great! I noticed a few out-of-date screenshots and added a couple comments, but it's all minor.

dsagal (approved):
Great, thanks!

anaisconce:
Looks really good! I like the little troubleshooting helpers a lot. I have a few comments.

1. Maybe add a couple lines to the intro to explain that formulas apply to the whole column and fields are included as $ColumnID. This information is already captured in `formulas.md`, but some users may only visit the cheat sheet. It's okay to borrow and repeat language here. 

2. For MAX or MIN example: The screenshot and formula uses lowercase Python syntax and the function is in uppercase spreadsheet function. This is okay, but maybe add a line to acknowledge that it's interchangeable.

3. The MAX example builds the list between parenthesis. The MIN example drills into a defined variable, and that variable is a list. The examples are great, but for novices, this difference might be hard to wrap their heads around. Maybe add a line to explain that in the MAX example, the list has two items: `$Max_Students - $Count` and `0`, and the formula returns whichever is greater. In the MIN example, the variable is pulling a list of records based on the lookupRecords arguments, listing the dates, and returning the smallest date (this is sort of explained but could be more explicit). 
Links to Record Set and dot notation article would be useful here, too.

4. For working with dates, let's add the `.strftime()` method. It come sup a lot. An example is in the general ledger template which has two hidden columns (Month and Year) that make use of `strftime()`. Can also link to this cheat sheet with the codes" https://strftime.org/

This was a lot of work! Thank you for compiling this, Natalie! 🙏🎉

nataliemisasi:
@anaisconce I did not include .strftime in this round but it is included on the Quip doc to be included in the next round of additions.

anaisconce (commented):
One last tiny comment. I should have been clearer last time about MIN v min.

anaisconce (approved):

PR #148, opened 2022-04-05
Add workflow action to run 'publish.sh live' on every commit to master

dsagal:
This should cause docs to be automatically published once a PR is landed to master. On balance, it seems less error-prone than requiring it as a separate manual step.

### DIFF
--- a/.github/workflows/publish-live.yml
+++ b/.github/workflows/publish-live.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - master
 jobs:
-  push-to-helpscout:
+  publish-live:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-live.yml
```diff
@@ -0,0 +1,21 @@
+name: Publish Live
+on:
+  push:
+    branches:
+      - master
+jobs:
+  push-to-helpscout:
```
paulfitz, line 7: doesn't really matter, but the job name is a bit misleading - copy/paste?

paulfitz (approved):

PR #152, opened 2022-04-18
Updated screenshots

nataliemisasi:
one of the formulas was incorrect and caused confusion for a user. In most of the formulas, it was written correctly but the screenshot was incorrect. Found one formula where it was incorrect and corrected that one.

WAS: user.Access != "owners" 
IS: user.Access != OWNER


dsagal (approved):
These are all helpful improvements -- thank you!

PR #156, opened 2022-05-25
Update Plugin API docs

georgegevoian:


paulfitz (approved):
Thanks @georgegevoian !

PR #157, opened 2022-05-27
Document new API endpoint for listing attachments

alexmojaki:
I've done my best to ensure I have the correct versions of all dependencies installed, but something must have changed.


api/grist.yml
```diff
@@ -543,6 +543,25 @@ paths:
               schema:
                 type: null
   /docs/{docId}/attachments:
+    get:
+      tags:
+        - /attachments
+      summary: "List metadata of all attachments in a doc"
+      parameters:
+        - $ref: '#/components/parameters/docIdPathParam'
+        - $ref: '#/components/parameters/tableIdPathParam'
```
berhalak, line 552: I think this is not required (the docs show it is)

berhalak (changes requested):

api/grist.yml
alexmojaki: oops, thanks

PR #162, opened 2022-07-26
update plugin documentation

paulfitz:
Main goal of this PR is to update `getAccessToken` documentation. But doc build was broken after typescript updates of main repo, so some tweaks were needed to get that going again.


berhalak (approved):
Thanks for the update.

PR #166, opened 2022-07-29
Fix push_to_helpscout script, and fix some image links

dsagal:
The `push_to_helpscout.py` script used HelpScout's [find redirect](https://developer.helpscout.com/docs-api/redirects/find/) API, which surprisingly does not return an ID that could be used to update that redirect. This PR switches it to use [list redirects](https://developer.helpscout.com/docs-api/redirects/list/) API, which does include that ID.

This was causing failures in the automatic workflow [Push to HelpScout](https://github.com/gristlabs/grist-help/actions/workflows/push-to-helpscout.yml).


georgegevoian (approved):
Thanks!

PR #172, opened 2022-08-29
give some guidance on system requirements for hosting Grist

paulfitz:
For simplicity, I stick with hosting Grist as a container. This could be a good place to collect requirements, rules of thumb for scaling, and known issues as we learn them.

### DIFF
--- a/help/self-managed.md
+++ b/help/self-managed.md
@@ -81,15 +81,15 @@ If using some other tool or service, here are the important points:
 Installed this way, Grist is accessible only to you. Typically you want to
 take at least the following steps:
 
-  * [Set up sandboxing](self-managed.md#sandboxing) - this is important to
+  * [Set up sandboxing](self-managed.md#how-do-i-sandbox-documents) - this is important to
     place bounds on what formulas can do.
-  * [Serve from a public host](self-managed.md#serving-from-a-public-host)
+  * [Serve from a public host](self-managed.md#how-do-i-run-grist-on-a-server)
     so you can collaborate live with others.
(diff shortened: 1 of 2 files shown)


help/self-managed.md
```diff
@@ -400,6 +400,46 @@ based on an export of the SendGrid templates for our SaaS.
...
+runs comfortably served from a Grist container with:
+
+  * 100MB RAM without sandboxing enabled.
+  * 200MB RAM with sandboxing enabled.
+  * 1 CPU.
+
+Memory and CPU requirements will scale with the number of documents
+simultaneously in use by your team.
+
+[Sandboxing](self-managed.md#sandboxing) is an important issue in serving
```
dsagal, line 429: This in-page link isn't working for me (I see at least one other place where it occurs from before this diff and also doesn't work)

dsagal (commented):
This is very useful. I found it informative too! Looks good, just something seems off about the anchor links.

help/self-managed.md
paulfitz: Oh good catch, lots of links broke when I rewrote section titles FAQ style... working on fixing.

paulfitz: I updated bad anchors in self managed docs. I think I've discovered others, will follow up in a separate PR.

dsagal (approved):
Thanks! I found two small typos unrelated to this diff:
- in line 227: "The UI elements present is also affected" -> "The UI elements present are also affected"
- in line 271: "you can made" -> "you can make".

PR #173, opened 2022-08-29
fix some internal links, and add a script to check them

paulfitz:
There are some internal links that have rusted, and aren't detected
as bad because only the anchor/hash is wrong, so the link reaches
a valid page (just not the right part of that page). This adds
a small script to apply the htmlproofer plugin to detect such
problems, and fixes a batch of them.

The plugin is a bit delicate. Headings in some files are migrated
from one markdown syntax to another to make the plugin happy.
I made a small fork of the plugin to make it deal better with
headers with style information. The plugin is not used by default,
and the fork is not installed by default, so as not to raise the
barrier of installing and using grist-help too much.

### DIFF
--- a/check_internal_links.sh
+++ b/check_internal_links.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+echo "Running htmlproofer plugin to check for internal link problems."
+
+if [[ "$1" = "install" ]]; then
+  ./env/bin/pip install -e git+https://github.com/gristlabs/mkdocs-htmlproofer-plugin@style-on-headings#egg=mkdocs-htmlproofer-plugin
+  echo "Installed."
+  exit 0
+fi
+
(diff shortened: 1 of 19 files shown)


check_internal_links.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+echo "Running htmlproofer plugin to check for internal link problems."
+
+if [[ "$1" = "install" ]]; then
+  ./env/bin/pip install -e git+https://github.com/gristlabs/mkdocs-htmlproofer-plugin@style-on-headings#egg=mkdocs-htmlproofer-plugin
+  echo "Installed."
```
dsagal, line 7: Maybe set `set -e`, so that the scripts doesn't claim success if pip-install fails?

dsagal (approved):
Thank you!

PR #176, opened 2022-09-07
answer some self-managed Grist questions (python packages, stored files)

paulfitz:
This answers some questions about how to install extra python packages (we don't have a good way to do this, but this consolidates answers elsewhere) and what files Grist stores in a persistent volume.


georgegevoian (approved):
Looks good to me. Thanks!

PR #179, opened 2022-09-14
add more metadata to pages

paulfitz:
This adds image and description metadata to pages, so previews in the forums are more useful.

 * A "description" meta-data field is used if available (I added just one as a test).
 * Otherwise a plugin tries to assemble a description from the first paragraph on the page.
 * Otherwise if the plugin can't figure out a description, the site description is used.

I needed some absolute urls, and `config.site_url` is apparently always modified to end in a `/`, so I used that - not sure if it is the best way.

### DIFF
--- a/overrides/base.html
+++ b/overrides/base.html
@@ -8,10 +8,19 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    {% if page and page.is_homepage %}<meta name="description" content="{{ config.site_description }}">{% endif %}
+    {% if page.meta and page.meta.description %}
+      <meta name="description" content="{{ page.meta.description|escape }}">
+      <meta property="og:description" content="{{ page.meta.description|escape }}">
+    {% else %}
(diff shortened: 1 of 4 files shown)


overrides/base.html
```diff
@@ -8,10 +8,19 @@
...
+    {% else %}
+      <meta name="description" content="{{ config.site_description|escape }}">
+      <meta property="og:description" content="{{ config.site_description|escape }}">
+    {% endif %}
     {% if config.site_author %}<meta name="author" content="{{ config.site_author }}">{% endif %}
     {% if page and page.canonical_url %}<link rel="canonical" href="{{ page.canonical_url }}">{% endif %}
     <link rel="shortcut icon" href="{{ '/images/favicon.png'|url }}">
+    <meta name="thumbnail" content="{{ config.site_url ~ 'images/favicon.png' }}">
+    <meta property="og:image" content="{{ config.site_url ~ 'images/favicon.png' }}">
+    <!-- {{config.site_url}} -->
```
georgegevoian, line 23:
```suggestion
```

georgegevoian (approved):
Works as expected. Thanks!

PR #182, opened 2022-09-15
add a summary of data security information

paulfitz:


anaisconce:
What do you think of having the article at the end of the menu as a topline item, not nested beneath "Reference"?

dsagal (approved):
Thank you! I think this is good. I don't feel it needs to be bumped up to top level of the table of contents. The important thing is that we can share a link to it (compare to pasting it into a a response, it's better to the user to see it as a public commitment), and also anyone searching for things like "security" or "privacy" will easily find it.

PR #184, opened 2022-09-28
Formula cheat sheet round 2

nataliemisasi:


help/formula-cheat-sheet.md
```diff
@@ -608,6 +608,9 @@ You can find duplicates in a column using either conditional formatting or a hel
 <span></span><details><summary>
 #### Example of Finding Duplicates
 </summary>
+
+**Video Walkthrough: [Finding Duplicates with a Formula](https://www.youtube.com/watch?v=7lNz2Mw7yyw&list=PL3Q9Tu1JOy_79QFrtQOEO1wHVIYYazdIf){:target="\_blank"}**
+
 **Community Example: [Finding Duplicates](https://public.getgrist.com/3CJkcpF7wu9Q/-1790/p/4){:target="\_blank"}**
 
 <span class="screenshot-large">*![duplicates-single-column](images/formula-cheat-sheet/duplicates-single-column.png)*</span>
```
anaisconce, line 616: I wonder if it would make sense to put all the external links after the formula and example, so it's optional to learn more, but not the primary action. The primary action is to learn the formula here. If the reader wants to learn more, there may be community threads, public example docs, and videos. 

So for all of these that have these "learn more" type of resources, maybe move it to after the example in the cheat sheet?

help/formula-cheat-sheet.md
```diff
@@ -862,3 +1017,224 @@ We can use this column to filter our data. If we only want to see interactions t
...
 </details>
+</section>
+
+<span></span><section class="cheat-sheet">
+#### Calculating the Difference Between Two Dates
+
+You can find the number of days between two [Date](col-types.md#date-columns) values using the [DAYS()](functions.md#days) function.
+
+<span></span><details><summary>
+#### Example of Calculating the Difference Between Two Dates
```
dsagal, line 1028: Sorry to not chime in earlier, but we support Excel-compatible function DATEDIF (https://support.getgrist.com/functions/#datedif). For days, it's exactly equivalent to what's here. For years or months, DATEDIF is much simpler, and I think we should recommend it in the cheatsheet instead of the more complicated solutions.

nataliemisasi:
I forgot to update Phone Number in this most recent commit. This is a reminder to myself to update it and re-commit.

PR #185, opened 2022-09-28
mention callback url in Auth0 SAML example; fix team name description

paulfitz:
This adds some clarifications that have come up:
  * The Auth0 SAML example didn't mention what the callback url should be.
  * The description of what GRIST_SINGLE_ORG can be set to was incorrect.

For `GRIST_SINGLE_ORG` a code fix is also coming to provide a better error message in case of problems.


dsagal (approved):
Ooh, this is much clearer.

PR #189, opened 2022-10-11
Upgrade redoc and update api docs

dsagal:
In particular, this fixes scrolling within api docs.


berhalak (approved):
Works ok. Thank you.

PR #191, opened 2022-10-25
Document new API endpoints for columns and tables

alexmojaki:
Documenting additions made in https://github.com/dsagal/grist/commit/3bdd35210df380859ca9049040fb586bf59bd5fa

The changes in `overrides/api-head.html` seem unrelated to my code changes, maybe something to do with local versions. I tried different node versions and reinstalling everything with pip and npm. Maybe someone else should rebuild.


berhalak:
I think api-head.html is always changing when we recreate API docs.

alexmojaki:
But it also changes when I build on master with no other changes: https://github.com/gristlabs/grist-help/compare/api-head-test

berhalak (approved):
Looks good to me. Regarding api-head.html, it also regenerates in master on my setup, so I think it is normal.

I think we need to **wait** with merge till Monday.

anaisconce:
Prod deployed! Merging to line up ducks for newsletter. Thank you both!

PR #194, opened 2022-11-02
Updates to screenshots for col-types

nataliemisasi:
updated screenshots in col-types article and linked to doc settings where applicable


dsagal (approved):
Thank you Natalie! These look good. One small thing I noticed -- there are two screenshots I noticed that show the column menu opened from the column dropdown arrow; the menu there is shifted up and overlaps the column header, so the arrow isn't visible. If there is a way to take a screenshot where the menu is just below the arrow, it would be clearer what to click to open that menu. If this is annoying to do, it's OK as is.

PR #198, opened 2022-12-07
Pinning filters

nataliemisasi:
Updated 'Filter Buttons' to reflect new pinning abilities.


help/integrators.md
```diff
@@ -8,6 +8,7 @@ include:
 - [Pabbly Connect](https://www.pabbly.com/connect/integrations/grist/)
 - [KonnectzIT](https://plan.konnectzit.com/feedback/grist-integration)
 - [n8n](https://n8n.io/integrations/n8n-nodes-base.grist)
+- [Make](https://www.make.com/en/integrations/grist) - Formerly Integromat
```
anaisconce, line 11: Ah, this change was in both this PR and the other PR to add make to integrator services. This one is no longer correct. Maybe just update it to reflect the change in the other PR.

help/search-sort-filter.md
```diff
@@ -143,37 +143,21 @@ As with sorting, you can create different widgets or pages showing the same tabl
 filter settings. For example, you could have one page showing event attendees with status
 "Confirmed" and another one showing those with status "Pending".
 
-### Filter Buttons
+### Pinning Filters
 
-Another way to apply filtering is through Filter Buttons. To use Filter Buttons, turn on the Filter Bar. 
-Click the “Sort & Filter” icon on the top-right of the table then select “Toggle Filter Bar”.
+Filters can be pinned at the top of a table for quick filtering. By default, filters will be pinned when first applied to a table. To unpin a filter, click the filter button then click the pin icon. When pinned, the pin icon will be green. When unpinned, the pin icon will be grey. If you unpin a filter button, any saved filters will still be applied but the button will no longer be visible.
```
anaisconce, line 148:
```suggestion
Filters can be pinned at the top of a table for quick filtering. By default, filters will be pinned when first applied to a table. To unpin a filter, click the filter button then click the pin icon. If you unpin a filter button, any saved filters will still be applied but the button will no longer be visible.
```
That's not quite right. Green means that the current filter is different from the saved filter. Gray means the current filter and saved filter are the same. So you can have a gray, pinned, and saved filter button.

help/search-sort-filter.md
```diff
@@ -143,37 +143,21 @@ As with sorting, you can create different widgets or pages showing the same tabl
...
-Filtering from the filter bar works exactly like the filtering described above. Note that there is a search bar at the top 
-that allows you to easily find values in longer lists.
+<span class="screenshot-large">*![pinning-filters](images/search-sort-filter/pinning-filters.png)*</span>
+{: .screenshot-half } 
 
-Click “Apply” to apply your settings.  
+If you want to save your pinned filters for future sessions, click the save button next to the “Search & Filter” icon.
 
-If you want to save your filters for future sessions, click the green checkmark next to the “Search & Filter” icon to save.
+The value in saving pinned filters with no filters applied is the ability to do quick filters and lookups without saving that filter forever, or for the rest of your team to see. So it may be useful to save the buttons without any filtering to create a quick filter toolbar.
```
anaisconce, line 155:
```suggestion
It may be useful to save the buttons without any filtering to create a quick filter toolbar.
```

I think just the one sentence is clearer. 🙂 WDYT?

anaisconce (changes requested):
Looks good overall! I made some small text edits. 

The way that the sort menu looks has also changed. May be good to update those screenshots as well.

anaisconce (approved):
Thank you!

PR #200, opened 2022-12-08
Adding fields description to /columns endpoint

berhalak:
Adding a schema and description for the `fields` parameter in the /columns endpoint. This should guide users on how to create and modify columns using this API.


api/grist.yml
```diff
@@ -1328,7 +1352,45 @@ components:
...
+          description: "A JSON object with widget options, e.g.: <code>{\"choices\": [\"cat\", \"dog\"], \"alignment\": \"right\"}</code>"
+        untieColIdFromLabel:
+          type: boolean
+          description: "Use \"true\" to indicate that the column label should not be used as the column identifier. Use \"false\" to use the label as the identifier."
+        recalcWhen:
+          type: integer
+          description: "A number indicating when the column should be recalculated. <ol><li>On new records or when any field in recalcDeps changes, it's a 'data-cleaning'.</li><li>Never.</li><li>Calculate on new records and on manual updates to any data field.</li></ol>"
+        recalcDeps:
+          type: string
+          description: "An encoded array of column identifiers (colRefs) that this column depends on. If any of these columns change, the column will be recalculated. E.g.: <code>[2, 3]</code>"
```
georgegevoian, line 1382:
```suggestion
          type: array
          description: "An encoded array of column identifiers (colRefs) that this column depends on. If any of these columns change, the column will be recalculated. E.g.: <code>["L", 2, 3]</code>"
```

georgegevoian (commented):
Thanks @berhalak. Just have one comment.

api/grist.yml
berhalak: Ahhh, thanks for noticing it. For POST and PATCH this is the correct description (we need an JSON encoded array of just numbers). But the GET endpoint returns the L versioned array.

georgegevoian: > Is it just me or does that feel weird?
> 
> I can see why something that returns a generic cell (that could have any kind of value in it) might do that. But colRefs are not arbitrary.

Yeah, it's definitely a little weird.

georgegevoian (approved):
Thanks @berhalak.

PR #201, opened 2022-12-22
update options for snapshots available for self-managed Grist

paulfitz:
There is now an option for using S3-compatible stores with Grist available in grist-core.


georgegevoian (approved):
Thanks @paulfitz!

PR #206, opened 2023-01-04
Removing experimental warning from ACL page

berhalak:
ACL rules are no longer an experimental feature.


paulfitz (approved):
:tada:

It would be good (but not vital) to delay landing this until after next prod release.

PR #207, opened 2023-01-11
Updates to acl banner

nataliemisasi:


nataliemisasi:
Memo updates included in this PR. If any more changes to memo PR, will update this one with those as well

anaisconce:
Thanks for updating the view as banner. There's a few other screenshots in the access rules page that use the older view as pill. Would it be easy to update those as well?

anaisconce (approved):
Thank you!

PR #209, opened 2023-01-27
Update screenshots for Formulas article, mainly to fix those broken in HelpScout beacon

dsagal:
HelpScout cached some images from this article in a broken way. To fix them, the images need a different checksum. Since many of them were out-of-date, I went ahead and created new screenshots. Made a little change to text to add a step about formatting.


nataliemisasi (approved):
Looks good 👍

PR #215, opened 2023-02-01
Duplicating Data

nataliemisasi:
adding 'Duplicating Data' section to Raw Data article


help/raw-data.md
```diff
@@ -4,18 +4,35 @@ The raw data page is a special page that lists all [data tables](glossary.md#dat
...
-![Raw Data View](images/raw-data-lightbox.png)
+![Raw Data View](images/raw-data/raw-data-lightbox.png)
+
+## Duplicating Data
+
+Tables can be duplicated from the Raw Data page. Click the three-dot icon to the right of the table you wish to duplicate then select 'Duplicate Table' from the menu.
+
+![raw-data-duplicate-table](images/raw-data/raw-data-duplicate-table.png)
+
+By default, a duplicated table will only contain the table structure, not the data. If you wish to copy all data in addition to the table structure, be sure to check the box prior to clicking 'Save'. If duplicating data and access rules exist for the original table, those rules **will not** be duplicated. Only the document default rules wil apply to the copy.
```
anaisconce, line 23:
```suggestion
By default, a duplicated table will only contain the table structure, not the data. If you wish to copy all data in addition to the table structure, be sure to check the box prior to clicking 'Save'. If the original table has any access rules, those rules **will not** be duplicated. Only the document's default rules will apply to the copied table.
```

help/raw-data.md
```diff
@@ -4,18 +4,35 @@ The raw data page is a special page that lists all [data tables](glossary.md#dat
...
+
+<span class="screenshot-large">*![raw-data-duplicate-data](images/raw-data/raw-data-duplicate-data.png)*</span>
+{: .screenshot-half } 
+
+This will create a new, unlinked duplicate table. Meaning, if you update data in this new copy, the original table will not be updated.
```
anaisconce, line 28:
```suggestion
The duplicate table is a new table that is not linked to the original. Meaning, if you update data in the copy, the original table will not be updated, and vice versa.
```

help/raw-data.md
```diff
@@ -4,18 +4,35 @@ The raw data page is a special page that lists all [data tables](glossary.md#dat
...
+By default, a duplicated table will only contain the table structure, not the data. If you wish to copy all data in addition to the table structure, be sure to check the box prior to clicking 'Save'. If duplicating data and access rules exist for the original table, those rules **will not** be duplicated. Only the document default rules wil apply to the copy.
+
+<span class="screenshot-large">*![raw-data-duplicate-data](images/raw-data/raw-data-duplicate-data.png)*</span>
+{: .screenshot-half } 
+
+This will create a new, unlinked duplicate table. Meaning, if you update data in this new copy, the original table will not be updated. 
+
+Note that instead of duplicating tables, it's usually better to segment data using [linked views](linking-widgets.md). For example, if you have multiple clients that all have similar data sets, you could keep all data for all clients in a single table but with the added column 'Client' to segment the data.
+
+***insert example as to when you would use this feature***
```
anaisconce, line 32: Oops I think you may have failed to commit some changes. :)

help/raw-data.md
```diff
@@ -4,18 +4,35 @@ The raw data page is a special page that lists all [data tables](glossary.md#dat
...
+![raw-data-duplicate-table](images/raw-data/raw-data-duplicate-table.png)
+
+By default, a duplicated table will only contain the table structure, not the data. If you wish to copy all data in addition to the table structure, be sure to check the box prior to clicking 'Save'. If duplicating data and access rules exist for the original table, those rules **will not** be duplicated. Only the document default rules wil apply to the copy.
+
+<span class="screenshot-large">*![raw-data-duplicate-data](images/raw-data/raw-data-duplicate-data.png)*</span>
+{: .screenshot-half } 
+
+This will create a new, unlinked duplicate table. Meaning, if you update data in this new copy, the original table will not be updated. 
+
+Note that instead of duplicating tables, it's usually better to segment data using [linked views](linking-widgets.md). For example, if you have multiple clients that all have similar data sets, you could keep all data for all clients in a single table but with the added column 'Client' to segment the data.
```
anaisconce, line 30: I'm a bit confused by how the "Client" column segments the data. Also not clear how to segment with linked views. 
 Maybe we can brainstorm a clearer example.

anaisconce (changes requested):
Thank you, Natalie! I made some comments.

anaisconce (changes requested):
Thanks! Lastly, `page-widgets.md` also has the `raw-data-nav.png` image. Moving it to another folder broke the link, which is why the test is failing. I think you'll need to fix the image location in the`page-widgets.md` page.

nataliemisasi:
Notes from discussion: update language around when *not* to use. Remove 'linking widgets' language. 

Example to use: Expenses - rather than having separate tables for each month, include all data in a single table with a new column, month. 

Mention that when you have a lot of similar columns to be used, the data should all be in the same (overlapping columns? make it one table)

No need to include an example as to when you would duplicate a table.

anaisconce (approved):
🙏

PR #216, opened 2023-02-06
Access Rules - Seed Rules

nataliemisasi:
Added info for new seed rule for owners and new schema edit rule


help/access-rules.md
```diff
@@ -43,29 +43,24 @@ the rules from top to bottom, and find the first applicable rule
 that allows (green) or denies (red) that permission.
 We'll see plenty of examples as we go.
 
+As we add more specific rules for tables, we typically want to grant OWNERs full access. Rather than adding a specific rule for each table to grant this access, we can check the box next to this option under Default Rules.
```
anaisconce, line 46: Let's add a header and call it "seed rules" like we do in product for consistency. 

Also, maybe this section can move down after "Make Private Table" and before "Restrict Access to Columns" so it's after the reader learns about specific table rules. It can be accompanied with an image of the seed rules uncollapsed showing the rule when the checkbox is toggled.

```suggestion
## Seed Rules
When writing access rules for specific tables, it is fairly common to repeat the same rule across many tables — for example, always grant owners full read and write permissions. To automatically add a set of rules to all new table rules, you can write "seed rules." There is a checkbox above default rules that makes the common case easier with one click. Click it to write a seed rule that will automatically grant owners full access whenever table rules are added. Click the `>` icon to uncollapse the seed rules table to modify seed rules. 
```

anaisconce (changes requested):
Made a comment and suggestion re: seed rules. Lock down structure looks good!

anaisconce (approved):
🙏

PR #219, opened 2023-02-08
Copy / Clone Document

nataliemisasi:
User had searched 'cloning document' and found nothing so contacted support. Adding text "clone" to copy document article so it will appear in searches.


anaisconce (approved):
Thank you!

PR #220, opened 2023-02-08
Removing Team Members

nataliemisasi:
adding section on removing team members and billing impacts


anaisconce:
Thanks for the edit! Text looks good. Could you crop the photo so there's less dead space? It'll help make the text and trash icon larger/clearer too.

anaisconce (changes requested):
Small change to the image requested in a comment above. 🙏

anaisconce (approved):
Thank you!

PR #221, opened 2023-02-13
highlight that the API needs a Content-Type of application/json

paulfitz:
A user had trouble using our API because they didn't realize Content-Type needed to be set - and it is easy to miss in the endpoint documentation.


berhalak (approved):
Thanks @paulfitz

PR #224, opened 2023-02-18
Skip steps of Github Actions that can't succeed, i.e. when secrets ar…

dsagal:
…en't available for PRs from public forks


dsagal:
I haven't actually taken the time to check that this will skip the problematic step on public forks, just relied on documentation in https://docs.github.com/en/actions/learn-github-actions/contexts#github-context.

paulfitz:
It looks plausible, let's try. If it doesn't work, we'll fix it :)

PR #226, opened 2023-02-20
add block for giving feedback on support pages

paulfitz:
Add a "thumbs up" / "thumbs down" area that feeds into logging.

### DIFF
--- a/overrides/base.html
+++ b/overrides/base.html
@@ -119,6 +119,50 @@
   {%- endif %}
 </div>
 
+<script>
+  function giveFeedback(event) {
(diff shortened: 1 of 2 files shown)


paulfitz:
@anaisconce does the thumbs up/down section in the preview look ok? It won't actually be functional until on *.getgrist.com. @georgegevoian I'm logging via an endpoint not built for this purpose, would like to replace it later with something this fits in better with other product feedback work.

georgegevoian (approved):
Code looks good. Thanks!

anaisconce:
🙏 Looks good!

PR #228, opened 2023-03-08
anonymous-access

nataliemisasi:
Updated 'manage team' screenshot to remove anon switch (does not exist), added anon section at end of team-sharing article


anaisconce (approved):
Thank you!

PR #232, opened 2023-03-29
there is no longer a postgres version limit

paulfitz:
Grist postgres support used to stop at version 11, but that is no longer the case.


georgegevoian (approved):
Thanks @paulfitz.

PR #240, opened 2023-05-10
remove inaccurate postgres version limit

paulfitz:
I missed some language about the postgres version limit in https://github.com/gristlabs/grist-help/pull/232


georgegevoian (approved):
Thanks @paulfitz.

PR #241, opened 2023-05-30
may-2023-newsletter

anaisconce:


help/newsletters/2023-05.md
```diff
@@ -0,0 +1,183 @@
...
+
+In March @CamilleLegeron made it possible to add descriptions to columns. Now, columns with descriptions have an info icon in their headers, which when hovered over or clicked reveal the respective description. This feature has been enabled on widgets as well, providing even more context to your team.
+
+To set a column or widget description, click into the title to open the renaming dialog which now includes a description field. Alternatively, descriptions may be set in the creator panel.
+
+![Column and Widget Description](../images/newsletters/2023-05/column-descriptions.png)
+
+### Webhooks!
+
+We're happy to release beta webhooks support! 🎉 Webhooks are one way to send information from Grist to other apps. For example, if a record is updated in Grist, send a message in Slack. When in a document, navigate to the Settings page listed the left-side panel under “Tools”. The “Manage Webhooks” button at the bottom will take you to the webhooks page.
```
dsagal, line 120: "For example, if a record is updated in Grist, send a message in Slack." -- I tried it, and I don't think this is possible directly (without an integrator service in the middle) because Slack expects a slightly different format than a Grist webhook can produce. I may have missed a way though, maybe @paulfitz knows?

help/newsletters/2023-05.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Expense Tracking for Teams
+
+Manage all employee expenses in one place. No more wrangling employees' expense spreadsheets into a master list every month. With access rules, employees can log into Grist, view and update only their expenses, and Grist takes care of the rest.
+
+Watch [this template’s webinar](https://www.youtube.com/watch?v=27sP9t-Q4m4){:target="\_blank"} to learn more about how this template was built.
+
+[![Expense Tracking for Teams](../images/newsletters/2022-06/expense-tracking.png)](https://templates.getgrist.com/55Q2EtTbFvB1/Expense-Tracking-for-Teams){:target="\_blank"}
+
+[GO TO TEMPLATE](https://templates.getgrist.com/55Q2EtTbFvB1/Expense-Tracking-for-Teams){:target="\_blank"}
```
dsagal, line 157: Usually our templates have some sample data, but this one has an empty Expenses table. Is that expected? (Not a problem, actually, just checking if intended.)

anaisconce: Access rules! The checkbox for fiddle mode strikes again. Must have made a small change at some point, and that box was unchecked. Should be fixed now.

help/newsletters/2023-05.md
dsagal: Maybe OK not to worry about it. If people attempt it directly and fail, we'll get some pressure for enhancements, but that would only be fair, since notifications to Slack are kind of a typical kind of thing to use webhooks for.

PR #243, opened 2023-06-08
grist.yml: fix schema for code generator

nafg:


nafg:
This makes https://guardrail.dev/ codegen work except for https://github.com/guardrail-dev/guardrail/issues/602 (which could be worked around with a simple `sed` etc.)

paulfitz (approved):
Thanks @nafg for the useful cleanup! Could you suggest a few lines we could add to our CI or run manually to check guardrail's happiness going forward? We'll be documenting more endpoints soon.

nafg:
@blast-hardcheese what is the easiest way to run the CLI in GitHub Actions?

blast-hardcheese:
@nafg Not sure how I missed this -- Easiest way would be to use coursier, `cs install --contrib guardrail` will give you a `guardrail` command on your $PATH that you can call directly.

Again, my apologies for not replying, I found this comment completely by accident when trawling open issues in the guardrail repo.

PR #244, opened 2023-06-09
document telemetry options for self-managed Grist

paulfitz:
This adds a first pass at documentation for telemetry options, and details of the data that would be sent. It will need revising once we have a UI for opting in to telemetry.

The data details are generated from the codebase, from `app/common/Telemetry.ts`. This is done to avoid accidental omissions, and to make it easier to keep up to date and accurate. The alternative would be speaking in generalities that don't need updating, which feels unsatisfying.

The `yarn cli settings telemetry` tool will be introduced in a pull request to a different repository.

### DIFF
--- a/build-telemetry.sh
+++ b/build-telemetry.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$1" = "" ]]; then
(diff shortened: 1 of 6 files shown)


georgegevoian (approved):
Looks good. Thanks!

PR #248, opened 2023-07-03
Column & Widget Descriptions

nataliemisasi:
Added bits for columns and widget descriptions


anaisconce (commented):
Looks good! As we discussed, let's add a "Renaming widgets" section beneath the "Changing a widget's type or its data". Symmetry with the renaming columns section, gets into renaming and descriptions.

anaisconce (approved):
Looks good. Thank you!

PR #249, opened 2023-07-06
OpenAPI: Fix the OrgKey mixed type

fflorent:
As per https://swagger.io/docs/specification/data-models/data-types/#mixed-type

The mixed types can be described using the combination of `oneOf` and a list of `type`s


dsagal (approved):
Thank you!

PR #250, opened 2023-07-07
Use the recommended Google Tag Manager snippets (both of them), and clarify the name of its identifier

dsagal:


dsagal:
I verified using Google Tag Manager debugging tool that GTM works now (it wasn't working previously), and since there are no content changes, I am merging without a review.

PR #261, opened 2023-08-21
Formatting-columns

nataliemisasi:
adding section on formatting cell and header styles


nbush:
Tiny changes (looks like we're almost always lowercasing "creator panel"), otherwise all good!

PR #264, opened 2023-08-23
import-dialog

nataliemisasi:
updated screenshots + text to reflect changes to import dialog


nbush:
`import-dialog-update-records.png` looks quite big (which I tried and failed to fix), but other than that this is good!

nbush (approved):

PR #273, opened 2023-10-11
Updates-to-Calendar

nataliemisasi:
updated article based on updates to Calendar since release


nbush (approved):
couple tiny changes but otherwise looks great!

PR #275, opened 2023-10-18
Document new download options and ability to use table ref instead of tableId

alexmojaki:
1. Documenting new options for downloading a document without history or data, as requested in https://community.getgrist.com/t/grist-document-download-options-in-the-api/3336
2. Updating the description of the common `tableId` path param to mention that numeric row IDs are also allowed now, as implemented in https://github.com/gristlabs/grist-core/pull/690

Also @paulfitz initially there were some changes in `api.md` related to `/sql` just from running `./api/build.sh` on a fresh checkout that are included here.


berhalak (approved):
Looks good, thank you.

PR #277, opened 2023-10-26
fix typo in col-refs.md

X1011:
correct the section name to that shown in the screenshot, also confirmed the correct name in production Grist UI

I don’t know what caused the phantom diff on the last line, but I’m guessing GitHub.com’s editor added a trailing newline.


dsagal (approved):
Thank you for the fix!

PR #280, opened 2023-10-31
add-column-menu

nataliemisasi:
updating to reflect the new column menu options


nbush (changes requested):
looks excellent aside from two very minor things:
- inconsistent use of quotation marks (i think we should stick to double for UI strings)
- when leading into the list of shortcut options, use a colon ("links below:")

nbush (approved):

PR #287, opened 2023-11-14
Issue 719 header param table schema doc

fflorent:
Document table-schema API.


api/grist.yml
```diff
@@ -492,13 +493,41 @@ paths:
...
+            description: "Name of a table (normalized)."
+          required: true
+        - $ref: '#/components/parameters/headerQueryParam'
+      responses:
+        200:
+          description: A table's table-schema in JSON format.
+          content:
+            text/json:
+              schema:
+                type: string
```
paulfitz, line 530: This return type isn't very informative?

paulfitz (commented):
Thanks @fflorent.

api/grist.yml
paulfitz: Just to be clear on what I mean. The generated documentation can render examples of responses on the side. For instance:

Screenshot from 2023-11-15 14-41-54

Currently, this PR adds an endpoint whose response is rendered like this:

Screenshot from 2023-11-15 14-42-18

fflorent: We could probably use this table-schema: https://specs.frictionlessdata.io/schemas/table-schema.json

fflorent: Done!

api/grist.yml
```diff
@@ -2074,6 +2103,37 @@ components:
...
+          type: string
+          enum: ["csv"]
+        mediatype:
+          type: string
+          enum: ["text/csv"]
+        encoding:
+          type: string
+          enum: ["utf-8"]
+        dialect:
+          $ref: https://specs.frictionlessdata.io/schemas/csv-dialect.json
```
fflorent, line 2133: If you feel uncomfortable with referencing external references, I can download the JSON and either include the YAML in the document or index as is locally in the repository. (also same remark for `schema`).

paulfitz (approved):
Great, thanks @fflorent !

PR #293, opened 2023-12-18
API spec cleanup: consistently use reusable path params and fix security scheme key

alexmojaki:
This is mainly about replacing repeated redefinitions of the orgId, workspaceId, and docId path params with refs to a single definition under components. As well as just making the spec shorter and cleaner, it makes it easier to build the customized API console.

In the params I've touched, I've moved the description from the schema up to the path definition because that's where it's actually supposed to be, otherwise it doesn't show in the Swagger UI. This should be done to all params, but I'll leave that to another PR.

Also I've fixed the `API Key` security scheme to not have a space because apparently that's bad.


paulfitz (approved):
Thanks for the clean-up @alexmojaki, good catch on the descriptions.

PR #294, opened 2023-12-26
Document Advanced Charts custom widget

alexmojaki:
Mostly just moved https://github.com/gristlabs/custom-charts-widget/blob/main/USAGE.md, but in particular removed the custom URL part.


paulfitz (approved):
(Given the new schedule, and @nataliemisasi being on vacation, I'm going to take this)

Included material has been reviewed elsewhere, and looks fine in its new home. Thanks @alexmojaki

PR #295, opened 2023-12-29
december-2023-newsletter

anaisconce:


help/newsletters/2023-12.md
```diff
@@ -0,0 +1,180 @@
...
+{: .grist-button}
+
+## What's New
+
+### Coming (very) soon: Forms
+
+![Coming soon...forms!](../images/newsletters/2023-12/forms-coming-soon.png)
+*Screenshot of the form editor's design.*
+
+Forms are almost here, and we couldn’t be more excited. We’ll release an early version for testing in January, and would love early feedback from users. To provide feedback, join our Discord server.
```
paulfitz, line 114: "early" used twice in same sentence - but far enough apart it isn't necessarily a problem? Maybe first early could be preview?

help/newsletters/2023-12.md
```diff
@@ -0,0 +1,180 @@
...
+
+[JOIN DISCORD](https://discord.gg/MYKpYQ3fbP){:target="\_blank"}
+{: .grist-button}
+
+### Beta Testing: Grist on AWS Marketplace
+
+![Grist coming soon to AWS marketplace](../images/newsletters/2023-12/aws-teaser.png)
```
paulfitz, line 121: ugh the image I gave you is small and blurry, I'll try and improve...

help/newsletters/2023-12.md
```diff
@@ -0,0 +1,180 @@
...
+      </td>
+    </tr></table>
+    <div class="header-welcome" style="color: #e0e0e0;">
+      Welcome to our monthly newsletter of updates and tips for Grist users.
+    </div>
+  </td>
+</tr></table>
+</div>
+
+![Grist 2023 Year in Review](../images/newsletters/2023-12/2023-year-in-review.png)
```
paulfitz, line 101: Like the banner, but the whitespace border around a full width banner like this feels slightly off... if it can be removed with a simple {:something} it'd be worth it, otherwise fine (not worth falling back on html)

anaisconce: Not that I know of. :( Actually, is there a place that lists all the {;somethings}? I've never found one.

paulfitz: Ok not to worry

The {: } are I think naming css classes which mostly come from here https://github.com/gristlabs/grist-help/blob/03f8af415677a47771d5011fc83ac76dc38f169c/help/css/grist.css#L116-L121

I expect they are irrelevant for the email (not sure of process there) in which case doubly don't worry

paulfitz (approved):
Thanks @anaisconce, looks great to me!

PR #300, opened 2024-01-10
add-column-menu-updates-and-glossary-additions

nataliemisasi:
updates to add column menu (col type) + glossary updates (lookups, trigger formulas)


anaisconce (approved):
Looks good, thank you!

PR #316, opened 2024-03-05
Fix hscout: Don't include MISSING-TRANSLATION.md in docs

fflorent:
Normally the path will still allow translating its content through Weblate.


mkdocs-utils/hooks.py
```diff
@@ -21,7 +21,7 @@
 
 @lru_cache
 def get_missing_translation_content(docs_dir: str) -> str:
-  missing_translation_file_path = Path(docs_dir) / "MISSING-TRANSLATION.md"
+  missing_translation_file_path = Path(docs_dir).parent / "MISSING-TRANSLATION.md"
```
CamilleLegeron, line 24: Maybe create a variable `MISSING-TRANSLATION` ? just a proposal because it's a string repeated many times

CamilleLegeron (approved):

paulfitz (approved):
Thanks @fflorent (and @CamilleLegeron !)

PR #317, opened 2024-03-18
Add French lang

fflorent:


paulfitz (approved):
Not sure I fully understand how translation is going to work, but I see that this starts making material available at /fr/ and doesn't affect existing material so I'm willing to take this step by step @fflorent :-)

fflorent:
@paulfitz Indeed, it just bootstrap a new language, and nothing apparent changes in the website unless a user adds a `/fr` in their URL

PR #320, opened 2024-03-26
Document the API payload limit

dsagal:


paulfitz (commented):
Could you remind me how this limit is controlled in the code, or the name of an environment variable for setting it? I flailed around ineffectually trying to confirm it.

dsagal:
Ah, good question. It's hard-coded here: https://github.com/gristlabs/grist-core/blob/a86c5da5c54b1caa49ac0971a6ef313b54a35df6/app/server/lib/FlexServer.ts#L904. An environment variable would be nice...

help/en/docs/limits.md
```diff
@@ -61,6 +61,10 @@ HTTP status code 429) until at least one of the original requests completes.  A
 for one request to complete before sending the next would not hit this limit (assuming it is the
 sole client accessing the document).
 
+The size of API requests is limited to 1MB. In particular, this applies to the total payload of
```
paulfitz, line 64: Wasn't entirely clear to me from the wording that this is a limit on the request body or the response to the request. "Total payload of API requests" feels ambiguous.

Maybe "The size of the body of any individual API request is limited..." ?

dsagal: Tried to rephrase for clarity.

PR #324, opened 2024-04-01
sort-by-function-change-quotes

MichaelNickolas:
updating the $ to quotes under the lookupone function


dsagal (approved):
Thanks Michael!

PR #325, opened 2024-04-01
Ignore __pycache__ directories, and removed a committed file there

dsagal:


paulfitz (approved):
omg thanks @dsagal - been seeing this junk recently and it was itching at me

PR #327, opened 2024-04-08
Add command to restrain language that can be added in weblate

fflorent:
The command is `./docs.py update-weblate-filters`

  Update the language filters (`language_regex` property), so only languages matching the regex can be added.
  Therefore no language can be added in Weblate unless they are present in the [alternate langs in the git repository](https://github.com/gristlabs/grist-help/blob/master/mkdocs-alternate-langs.yml).

This means that we may allow translator create languages for any component, we control what are those we allow to be created through this filter.

  Also see:
   - https://github.com/WeblateOrg/weblate/discussions/9703
   - https://docs.weblate.org/en/latest/admin/projects.html#component-language-regex

Demo in Weblate:

https://github.com/gristlabs/grist-help/assets/371705/42272d6e-f01f-4cf5-8247-e912678122bf

### DIFF
--- a/docs.py
+++ b/docs.py
@@ -10,11 +10,14 @@ import logging
 import os
 import shutil
 import subprocess
+import requests
+from getpass import getpass
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 from multiprocessing import Pool
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 from textwrap import dedent, indent
+from requests.adapters import HTTPAdapter, Retry
 
 import mkdocs.commands.build
 import mkdocs.commands.serve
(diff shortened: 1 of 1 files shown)


docs.py
```diff
@@ -174,6 +178,53 @@ def update_languages() -> None:
...
+
+  Also see:
+   - https://github.com/WeblateOrg/weblate/discussions/9703
+   - https://docs.weblate.org/en/latest/admin/projects.html#component-language-regex
+  """
+
+  api_key = os.environ.get("WEBLATE_API_KEY")
+  grist_help_project_slug = 'grist-help'
+  if not api_key:
+    api_key = getpass("Enter Weblate API Key: ")
```
paulfitz, line 195: Maybe say `(or set WEBLATE_API_KEY in environment)` here or in an earlier line? So a user who cares will know about that option.

paulfitz (approved):
Looks fairly straightforward, and low stakes if it proves to not be quite right.

Thanks @fflorent

paulfitz:
@fflorent I believe you have privileges to land this when you're happy with it (now that it is approved).

fflorent:
Thanks @paulfitz for your review!

PR #330, opened 2024-04-12
list some reasons why gvisor sandboxing may fail

paulfitz:
This follows a suggestion given in https://github.com/gristlabs/grist-core/issues/914#issuecomment-2022610020

### DIFF
--- a/help/en/docs/self-managed.md
+++ b/help/en/docs/self-managed.md
@@ -124,12 +124,33 @@ import glob
 glob.glob('/etc/*')
 ```
 
-There are some reasons why `gvisor` sandboxing, as configured for
-Grist, may fail.
+Here are some reasons why `gvisor` sandboxing, as configured for
+Grist, may fail, and what you can do to diagnose the problem.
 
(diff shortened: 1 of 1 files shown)


help/en/docs/self-managed.md
```diff
@@ -124,6 +124,13 @@ import glob
 glob.glob('/etc/*')
 ```
 
+There are some reasons why `gvisor` sandboxing, as configured for
+Grist, may fail.
+
+ * Your processor may not be supported. On `x86_64`, Sandy Bridge
+   or later is needed. Check that the `XSAVE` processor flag is set.
+ * The `SYS_PTRACE` capability may not be available.
```
jordigh, line 132: We probably should give some indication of how to check these two.

We're talking about Linux here, so we can tell people something like `if grep -q '\bxsave\b' /proc/cpuinfo; then echo "XSAVE enabled"; else echo 'XSAVE not present'; fi;` or doing `--cap-add=SYS_PTRACE ` for `docker run`.

jordigh (changes requested):

jordigh (approved):

PR #333, opened 2024-04-26
make.com-urlchange-textdeletion

MichaelNickolas:
Make.com requiring changes to the URL and the elimination of "formerly Integromat"


dsagal (approved):
Thanks!

PR #338, opened 2024-04-30
self-managed: explain how to configure websockets

jordigh:


paulfitz (approved):
Perfect, that's great, thank you @jordigh !

PR #340, opened 2024-05-07
remove user.Origin

paulfitz:
This was a feature that was never fleshed out, which has caused trouble, and which will shortly be removed: https://github.com/gristlabs/grist-core/pull/915

I propose removing the documentation for the feature entirely. Another option would be to strike it out and link to the discussion above. But given that we've no evidence anyone is using it, I've just deleted it. If someone complains we can revisit.


jordigh (approved):

fflorent (approved):
Thanks @paulfitz

PR #343, opened 2024-05-30
resizing images

nataliemisasi:


nataliemisasi:
confirmed sizing looks correct. Publishing so live article looks better 👍

PR #348, opened 2024-05-31
correcting title to 'Formula timer'

nataliemisasi:


nataliemisasi:
minor change - Nick pointed out an inconsistency in the title within navigation
image

PR #350, opened 2024-06-10
updating limits to include data size

nbush:


help/en/docs/limits.md
```diff
@@ -33,24 +33,27 @@ table with the product type as an extra column.
...
-also on the number of tables, columns, and the average size of data in each cell. One way to
-estimate it is to measure the size of the data when it is in CSV format: the limit is around 20MB
-in this format. For example, a document with 200,000 rows and 12 numeric columns would reach that.
+Documents are also subject to data size limits, as described below.
+
+## Data size
+
+There is a hard limit to a document's total data size, determined as the row limit multiplied by 2KB. This means that documents on the Free plan have a data size limit of 10MB, with Pro and Business plan documents having limits of 200MB and 300MB, respectively. This value corresponds approximately to the size of the data in CSV format. You can see a document's current data size on the ['Raw Data' page](https://support.getgrist.com/raw-data/#usage).
+
+We recommend 20MB as a soft limit for optimal performance, as documents beyond 20MB in data size may slow down. To help optimize formulas on large documents, you can use the built-in [formula timer](https://support.getgrist.com/formula-timer/).
```
dsagal, line 44: Sorry, I forgot to say an important thing: the 20MB limit, previously described as a "rule of thumb" is not only about performance but also about whether it risks exceeding memory limits and crashing. Memory limits are harder to talk about, or predict, and we can't really want to make promises at all. E.g. a single formula could do some dumb Python stuff that uses too much memory and causes the document to crash (leading to "PipeToSandbox" we see reported occasionally). But also a large document -- especially with lookup formulas -- will be more likely to hit memory issues. (On the other hand, improvements we make internally may reduce such memory pressures in the future.) So it's better to call it maybe a "recommended limit", and mention the risk of a document not working at all (data engine crashing?) being one reason, and performance being the secondary one.

Everything else looks great! Thanks Nick!

nbush: How about this?

> For memory and performance reasons, there's a recommended data size limit of 20MB. Documents beyond 20MB may slow down or run into memory limits depending on their complexity. To help optimize formulas on large documents, you can use the built-in [formula timer](https://support.getgrist.com/formula-timer/).

dsagal: That's clear, thanks!

Do you think it would help to keep some of the bits from the previous version to clarify the relationship with row limits? E.g. if we added an example like this the first sentence?

> For memory and performance reasons, there's a recommended data size limit of 20MB. For example, a document with 100,000 rows and 24 numeric columns would reach that. Documents beyond 20MB may ...

nbush: Done!

dsagal (approved):
Thank you!

PR #352, opened 2024-06-13
table-wide-image-change

MichaelNickolas:
made changes to the images and table-wide rule wording


MichaelNickolas:
I will make full article formatting changes in a new PR once we approve the image changes.

nataliemisasi (approved):
looks great

nbush (changes requested):
I think we should establish a practice of deleting unused images so that the dir doesn't get too cluttered. Otherwise looks good!

MichaelNickolas:
> I think we should establish a practice of deleting unused images so that the dir doesn't get too cluttered. Otherwise looks good!

There are no images in the images folder that are not being used.

PR #353, opened 2024-06-17
record card layout text addition

MichaelNickolas:
Layouts saved for record cards... text addition


nataliemisasi (changes requested):
Per discussion in Slack w/ Nick, move the note from the end of this section to the beginning. Update start of section to the following:

"You can edit a record card’s default layout from the Raw Data page. This new layout will apply to any new card widgets using the same table as source data. Click the card icon to open."

nataliemisasi (approved):

PR #355, opened 2024-06-20
[Tiny] Document missing MINIO env variables

fflorent:
There are two environment variables that are not documented for MinIO. This PR fixes that.

Fixes https://github.com/gristlabs/grist-core/issues/1058


paulfitz (approved):
Wonderful, thank you @fflorent

PR #360, opened 2024-07-09
Show language selector!

fflorent:
This allows the user to show the documentation in French for now and in other languages in the future.

### DIFF
--- a/help/en/MISSING-TRANSLATION.md
+++ b/help/en/MISSING-TRANSLATION.md
@@ -1,3 +1,5 @@
-The current page still doesn't have a translation for this language.
+We're just getting started translating this language, sorry!
 
-But you can help translating it: [Contributing](https://hosted.weblate.org/engage/grist-help/){.internal-link target=_blank}.
+We show partially translated languages to track progress.
+
+This page isn't translated yet. But the good news is that [you can join the translation community to help us 👋](https://hosted.weblate.org/engage/grist-help/){.internal-link target=_blank}.
--- a/help/en/docs/css/grist.css
+++ b/help/en/docs/css/grist.css
@@ -175,7 +175,7 @@
 .btn-group.wm-vcentered { display: inline-flex; }
 .btn-group.wm-vcentered > .btn.btn-sm { float: none; }
 
-.home-link {
+.home-link, .home-top-tool-selector {
   font-size: 18px;
   margin-right: 3rem;
   flex: 0 0.1 auto;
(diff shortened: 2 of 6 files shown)


paulfitz (commented):
Hmm, once this happens, the selector will show on support.getgrist.com. The French translation is so bare, including on the default/front page, that it might be mostly disappointing for users.

If you don't care about support.getgrist.com, we could maybe add a flag so the selector doesn't show their yet. But if getting it to show there is the point, I can check in with Grist Labs support people to see how they feel about this.

The French translation is already available at https://support.getgrist.com/fr/

fflorent:
I expected that it would encourage to help translating the website, if we advertise it as needing some work from the community.

That being said, probably in such a case we should either make it explicit that it's still in beta phase, or continue translating on our side before making it public.

I don't mind waiting until we have more content in French (so opt for the second option), and until then close or make this PR as "WIP".

fflorent:
@paulfitz That's interesting. I would need to be sure of the purpose(s) you have in mind for the message so I can enhance relevantly (also I may need to tweak a bit the code depending on your answer):
 - advertise the new localization and notify that it's in very early stage (that would indeed help);
 - help existing translators to track the progress as you mentioned (for that, we may redirect them to https://support.getgrist.com);
 - encourage the community to help translating;
 - other purposes?

vviers:
I like the option of having the "work in progress, help needed" warning message @paulfitz ! Sorry for falling behind on translation we'll get to it this summer !

fflorent:
I propose this message based on your suggestion:

> We're just getting started translating this language, sorry!
> 
> We show partially translated languages to track progress.
> 
> This page isn't translated yet. But the good news is that [you can join the translation community to help us 👋](https://hosted.weblate.org/engage/grist-help/){.internal-link target=_blank}.

It's rendered like below in French:

image

paulfitz (approved):
Thank you, the revised language works well! Checked in with rest of support team, they are excited for the French translation to make progress. in any way. One "nice to have" that came up would be to match style on the language dropdown with the style used in the main application. But for the moment a less polished and less consistent look may actually set expectations more appropriately.

Leaving it to you to land at your convenience @fflorent.

fflorent:
@paulfitz Do you mean like this? (The font size is set like for the other elements on the top)
image

fflorent:
I pushed an update of the code in a separate commit. Your feedback is welcome

paulfitz:
Thanks a welcome tweak. But what we hand in our mind with this kind of dropdown used in the Grist app:

language-dropdown

This PR is good to land in my opinion, and I wouldn't actually want to hold it for a discussion about a fancier dropdown. We could revisit when there's at least one complete translation :-)

fflorent:
Thanks for the clarification!

I merge as suggested, we'll work on that in a follow-up

PR #362, opened 2024-07-15
References updates

nataliemisasi:


help/en/docs/col-refs.md
```diff
@@ -180,38 +180,38 @@ We can see that the value in the **Full Name** column for the record with Row ID
...
 
-*![Unfiltered reference dropdown list](images/filter-reference-columns/unfiltered-cities.png)*
+*![Unfiltered reference dropdown list](images/columns/unfiltered-cities.png)*
 {: .screenshot-half } 
 
-It would be useful if the dropdown list of city choices were filtered based on the country selected in the **Country** column.
+It would be useful if the dropdown list of city choices were filtered based on the country selected in the **Country: Pop Ranking** column.
 
-To filter a reference column’s dropdown list, select the reference column then set a 'Dropdown Condition' in the *creator panel* under the 'Column' tab. 
+To filter a reference column’s dropdown list, select the reference column then 'Set dropdown condition` in the *creator panel* under the 'Column' tab.
```
nbush, line 190: Closing backtick ` instead of '.

help/en/docs/col-refs.md
```diff
@@ -180,38 +180,38 @@ We can see that the value in the **Full Name** column for the record with Row ID
...
-To filter a reference column’s dropdown list, select the reference column then set a 'Dropdown Condition' in the *creator panel* under the 'Column' tab. 
+To filter a reference column’s dropdown list, select the reference column then 'Set dropdown condition` in the *creator panel* under the 'Column' tab. 
 
-*![Set dropdown condition](images/filter-reference-columns/set-dropdown-condition.png)*
+*![Set dropdown condition](images/columns/set-dropdown-condition.png)*
 {: .screenshot-half } 
 
 
-You can filter a dropdown’s choice by writing a condition as a formula. The attribute `choice` refers to choices in the dropdown. In this case the formula is `choice.Country == $Country`. 
+You can filter a dropdown’s choice by writing a condition as a formula. The attribute `choice` refers to choices in the dropdown. In this case the formula is `choice.Country_Cities == $Country_Pop_Ranking`.
```
nbush, line 196: I feel like this is where it gets a bit confusing for me, because I have no idea of the structure of the overall document. I fear the colons make the confusion a worse, as they are kind of pseudo-hierarchical and get removed when put in a formula.

A drastic suggestion if you think it may help:
1. Drop the population ranking framing
2. Two tables, Filtered Cities & Raw Cities (or similar)
3. In Raw Cities, have **Country** and **City** columns
4. In Population Rankings, have **Filter Country** and then just **Filtered City**

In this case the formula should be `choice.Country == $Filter_Country`, which should hopefully eliminate the ambiguity, unless I've screwed something up.

nbush (changes requested):
One somewhat major suggestion that I think might help simplify things, but open to other ideas.

nbush (changes requested):
One final request to add a link to the Country/City doc in the opening paragraph of the section to show the reference structure.

jr-grist (changes requested):
Explain structure of tables in example before diving into the formula and introducing choice.Country

nataliemisasi:
merging to add current changes. Jordan will create a separate PR to add link and further explanation

PR #367, opened 2024-07-29
documentation: user deletion API

fflorent:
I see that there is an endpoint for deleting users, but it is undocumented.

Still this endpoint can be useful, and I am interested in it (also in evolving it). I am documenting for now its current behavior, if you think it is ready to be officially proposed.


api/grist.yml
```diff
@@ -1124,6 +1157,10 @@ tags:
     description: "Documents may include attached files. Data records can refer to these using a column of type `Attachments`."
   - name: webhooks
     description: "Document changes can trigger requests to URLs called webhooks."
+  - name: sql
+    description: "Sql endpoint to query data from documents."
+  - name: users
+    description: "Users of Grist"
```
vviers, line 1163:
```suggestion
    description: "Grist users."
```

api/grist.yml
```diff
@@ -1104,6 +1104,39 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/SqlResultSet"
+  /users/{userId}:
+    delete:
+      tags:
+        - users
+      summary: "Delete a user from Grist"
+      description: |
+        Also delete the user's personal org (aka team site).
```
vviers, line 1113:
```suggestion
        This action also deletes the user's personal organisation and all the workspaces and documents it contains.
```

api/grist.yml
```diff
@@ -1104,6 +1104,39 @@ paths:
...
+        Currently only the user themselves is allowed to delete their own account.
```
vviers, line 1114:
```suggestion
        Currently, only the users themselves are allowed to delete their own accounts.
```

api/grist.yml
```diff
@@ -1104,6 +1104,39 @@ paths:
...
+
+        ⚠️ **This cannot be undone, please be cautious when using this endpoint** ⚠️
```
vviers, line 1116:
```suggestion
        ⚠️ **This action cannot be undone, please be cautious when using this endpoint** ⚠️
```

api/grist.yml
```diff
@@ -1104,6 +1104,39 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - name
+              properties:
+                name:
+                  type: string
+                  description: "The user's name to delete (for confirmation, to avoid deleting the wrong account)."
```
vviers, line 1129: Not an issue, but intuitively I would think that passing an email address would be easier than passing a user name 🤷🏼

vviers (approved):
LGTM, minor edits suggested

api/grist.yml
fflorent: Good point, especially the normalized one. Though if we take a look at the [database structure](https://github.com/gristlabs/grist-core/blob/main/documentation/database.md#the-schema), the email is part of the *logins* table. Though there should only be a user per email address, so I feel like this should be OK.

Reported here: https://github.com/gristlabs/grist-core/issues/1134

paulfitz (approved):
Thanks you @fflorent !

PR #368, opened 2024-07-30
Generate documentation for PREVIOUS/NEXT/RANK functions, .find.* methods, and related updates to lookups

dsagal:


help/en/docs/functions.md
```diff
@@ -21,14 +21,15 @@ Python (see [Python documentation](https://docs.python.org/3.11/)). Here are som
...
+| Prevnext | <a  href="#next">NEXT</a>, <a  href="#previous">PREVIOUS</a>, <a  href="#rank">RANK</a> |
```
nbush, line 30: Can we use "Cumulative" here instead? This is a bit odd with RANK sitting right there.

help/en/docs/functions.md
```diff
@@ -103,6 +104,54 @@ min(Tasks.lookupRecords(Owner="Bob").DueDate)
...
+
+The methods available are:
+
+- __`lt`__: (less than) find nearest record with sort values < the given values
+- __`le`__: (less than or equal to) find nearest record with sort values <= the given values
+- __`gt`__: (greater than) find nearest record with sort values > the given values
+- __`ge`__: (greater than or equal to) find nearest record with sort values >= the given values
+- __`eq`__: (equal to) find nearest record with sort values == the given values
+
+Example from <https://templates.getgrist.com/5pHLanQNThxk/Payroll>. Each person has a history of
```
nbush, line 129: Suggest "Example from [our Payroll template](https://templates.getgrist.com/5pHLanQNThxk/Payroll)."

help/en/docs/functions.md
```diff
@@ -3330,6 +3423,75 @@ This would only calculate UUID() once and freeze the calculated value. By contra
...
+The `order` parameter may be `"asc"` (which is the default) or `"desc"`.
+
+When `order` is `"asc"` or omitted, the first record in the group in the sorted order would have
+the rank of 1. When `order` is `"desc"`, the last record in the sorted order would have the rank
+of 1.
+
+If there are multiple groups, there will be multiple records with the same rank. In particular,
+each group will have a record with rank 1.
+
+For example, `RANK(rec, group_by="Year", order_by="Score", order="desc")` will return the rank of
```
nbush, line 3491: > For example, `RANK(rec, group_by="Year", order_by="Score", order="desc")` will return the rank of
> the current record (`rec`) among all the records in its table for the same year, ordered by
> decreasing score.

Potentially dumb question: can you achieve the same thing as the `order` parameter by just adding/removing `-` to `order_by` (`order_by="-Score"`)?

nbush (changes requested):
A couple tiny questions/changes requested.

help/en/docs/functions.md
dsagal: Yes. If you reuse the same `order_by` settings for multiple purposes, that's slightly more efficient, but really no reason to get into that. Just more than one way to do the same thing; similar to why `NEXT` exists.

nbush (approved):

PR #369, opened 2024-07-31
API: /docs/{docId}/move may move to a WS of a different org

fflorent:
## Context

While looking at the API documentation, I read that the `PATCH /docs/{docId}/move` endpoint specifies that the doc may move to a workspace at the condition that it remains in the same org.

We would like to use this endpoint to move the document of an old account's personal org to a new account's one.

After having taken a look [at the code](https://github.com/gristlabs/grist-core/blob/42f696d50f01d35b43b413295a2977497c94480d/app/gen-server/lib/homedb/HomeDBManager.ts#L2312-L2326) and experimented, it looks like it is not the case.

## Proposed solution

I simply remove the mention indicating that the target workspace should be in the same org as the source.


paulfitz (approved):
Thanks! The summary does seem to be simply wrong.

PR #381, opened 2024-08-08
self-hosted: document that there is a known problem with postgres

jordigh:
This is a workaround for https://github.com/gristlabs/grist-core/issues/773

### DIFF
--- a/help/en/docs/self-managed.md
+++ b/help/en/docs/self-managed.md
@@ -633,8 +633,23 @@ and set the following variables:
 * TYPEORM_HOST - set to hostname of database, e.g. grist.mumble.rds.amazonaws.com
 * TYPEORM_PORT - set to port number of database if not the default for PostgreSQL
 
-Grist is known to work with PostgreSQL from versions 10 through 14 (later versions
-are likely to work also, but have not been specifically tested at the time of writing).
+Grist is known to work with PostgreSQL from versions 10 through 16.
+Recent versions, however, have enabled by default a JIT compiler that
+is known to cause problems with Grist, which expresses itself as every
+cell operation taking a few noticeable seconds. In case this happens,
+PostgreSQL's JIT compiler should be disabled for Grist with the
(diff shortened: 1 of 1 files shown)


help/en/docs/self-managed.md
```diff
@@ -627,8 +627,19 @@ and set the following variables:
 * TYPEORM_HOST - set to hostname of database, e.g. grist.mumble.rds.amazonaws.com
 * TYPEORM_PORT - set to port number of database if not the default for PostgreSQL
 
-Grist is known to work with PostgreSQL from versions 10 through 14 (later versions
-are likely to work also, but have not been specifically tested at the time of writing).
+Grist is known to work with PostgreSQL from versions 10 through 16.
+Recent versions, however, have enabled a JIT compiler that is known to
+cause problems with Grist. This JIT compiler should be disabled for
```
paulfitz, line 632: I was interested in the situation for major providers of postgres as a service. For AWS, the only mention I found was this https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-versions.html where in the introduction of version 11 a way to enable JIT if you want it is noted. I'm guessing JIT remains off by default, since we've never hit this problem.

I wondered also about the default debian package for postgres, what it does, since again I'm pretty sure (but not 100% sure) we tested against this.

I don't know if there's anything to add here except perhaps to be less definitive about the "have enabled" since depending on how the user is getting their postgres, JIT may or may not be enabled. If it turns out the cloud providers are keeping the flag off, for example, then it would be a shame to have installers waste their time unnecessarily. Maybe give the symptom to watch out for?

jordigh: Hm, it's enabled by default in Debian.
image
I'll reword this a bit to be a bit more vague about how to disable JIT.

jordigh: Rewording done. How does it look now?

dsagal (approved):
Thanks Jordi!

PR #383, opened 2024-08-13
Doc api keys

hexaltation:
Fixes gristlabs/grist-core#579

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Repository for Grist documentation and tutorials.
 
 The public site is at <https://support.getgrist.com/>, served using GitHub Pages.
 
-The content is hosted at <https://github.com/dsagal/grist-help/>. On changes to master (such as
+The content is hosted at <https://github.com/gristlabs/grist-help/>. On changes to master (such as
 when a pull request is merged), the public site gets automatically republished.
 
 Previews of pull requests are automatically built by Netlify, and made available at URLs such as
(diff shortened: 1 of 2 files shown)


fflorent:
> Fixes https://github.com/gristlabs/grist-core/issues/579

Wait, will it really fix this issue?
Using this syntax will close as completed the above issue.

hexaltation:
> > Fixes [gristlabs/grist-core#579](https://github.com/gristlabs/grist-core/issues/579)
> 
> Wait, will it really fix this issue? Using this syntax will close as completed the above issue.

I'm ok that it's not perfect but I want to keep track the fact that this PR needs to be merge at the same time as the functional one.

`/api/grist.yml` is much code than doc to me, as it's functionally used in production by console api to interact with the app.

Can't it be better to move it to gristlabs/grist-core and to pull it when doc is built ?

@gristlabs people do you have an opinion over this ?

dsagal:
I agree with @fflorent that the comment on the PR seems risky -- not sure it would actually close the issue in the other repo. Anyway, it would be helpful to have a clearer title and description for the PR, perhaps like "Add OpenAPI spec for the new Doc API keys endpoints in https://github.com/gristlabs/grist-core/issues/579"?

I also agree with @hexaltation suggestion to move `/api/grist.yml` into `grist-core`! Good idea! It would make it easier to keep `grist.yml` up-to-date. It would also help keep the API Console in sync with the current version of Grist.

PR #386, opened 2024-08-15
Markdown version upgrade now requires a 'markdown' attribute for html elements that contain markdown

dsagal:


nbush (approved):
Looks good!

PR #384, opened 2024-08-14
From weblate to gitlocalize

fflorent:
This PR:
 - removes any mention of weblate in favor of gitlocalize
 - removes a function related to weblate in the `docs.py` script
 - warns about the translation being generated
 - allow people asking to help translating through github using the dedicated template. Regular issues can still be open. [Preview available through my fork](https://github.com/fflorent/grist-help/issues/new/choose).

Fixes #370

PR #388, opened 2024-08-16
Machine translation using OpenAI

fflorent:
## Use GPT-4o and GPT-4o-mini to machine translate the documentation

This PR adds translations made using GPT-4o for pages who are below 4096 tokens length.

I wrote and used [this script](https://gist.github.com/fflorent/96c057da668b2d0e1c4b924f882e9448) to translate the documents into French. It is not well documented and purely written to be run in one-shot, I can give more details on how it works upon request.

One notable thing I used is [a glossary](https://gist.github.com/fflorent/96c057da668b2d0e1c4b924f882e9448#file-batch-py-L22-L40) that I gave to GPT through the prompt. It looks like GPT-4o follows it much more carefully than GPT-4o-mini.

Related PR: #384, which adds the headers and templates to move from Git

**Important**: some pages are not translated yet. That's normal, I still want to see whether Git Localize achieve synchronizing many documents like the set in this PR, I will continue translating after that.


fflorent:
Waiting for merge of #384 before merging this PR.

PR #391, opened 2024-08-22
oidc: clarify a small point of quoting an envvar

jordigh:
Ran into this problem by quoting in my `.env` file for Docker:

https://community.auth0.com/t/id-token-not-present-in-tokenset/63690/2


help/en/docs/install/oidc.md
```diff
@@ -84,6 +84,6 @@ Then, you can start Grist with the following configuration:
 
  - `GRIST_OIDC_SP_HOST` = `https://<grist-domain>` / `http://localhost:8484`
  - `GRIST_OIDC_IDP_ISSUER` = `https://<keycloak-domain>/realms/<your-realm>`
- - `GRIST_OIDC_IDP_SCOPES` = `"openid profile email"`
+ - `GRIST_OIDC_IDP_SCOPES` = `"openid profile email"` (if using an `.env` file or similar, make sure to omit the quotes, which are only needed when defining this variable from a shell)
```
dsagal, line 87: The same section occurs several times in the file, but I am not sure about repeating the long instruction in each.

How do people usually use it? If it's with `.env` file, would it be clearer to list a single code block in .env file format (and comments after "#"), and add a note below the block like "This format is suitable for an .env file. For shell, remember put quotes around any values with spaces"?

jordigh: I thin it's okay to repeat stuff as people are likely to just click to the section for their particular example and not read the other sections. Redundancy is good for humans, less good for code maintenance. :)

I've made some changes based on your recommendation.

dsagal: Could you check the preview? Something seems off with formatting (maybe the closing code-block quotes have an extra space in front?)

help/en/docs/install/oidc.md
```diff
@@ -44,12 +44,19 @@ See [how to create an OAuth2 application in Gitlab in this documentation](https:
...
- - `GRIST_OIDC_IDP_ISSUER` = `https://gitlab.com/.well-known/openid-configuration`
- - `GRIST_OIDC_IDP_SCOPES` = `"openid profile email"`
- - `GRIST_OIDC_IDP_CLIENT_ID` = "..." (the client ID generated by Gitlab for the application)
- - `GRIST_OIDC_IDP_CLIENT_SECRET` = "..." (the client secret generated by Gitlab for the application)
- - `GRIST_OIDC_IDP_SKIP_END_SESSION_ENDPOINT` = `true` (Gitlab don't propose `end_session_endpoint`)
+```shell
+GRIST_OIDC_SP_HOST=https://<grist-domain> # or http://localhost:8484
+GRIST_OIDC_IDP_ISSUER=https://gitlab.com/.well-known/openid-configuration
+GRIST_OIDC_IDP_SCOPES=openid profile email
+# the client ID generated by Gitlab for the application
```
fflorent, line 51: Nitpicking: I would add colons (`:`) at the end of the line of this kind of comment to tell it relates to the next line.
It took me (few) seconds to realize that.

fflorent (approved):
I am sorry you ran into problems @jordigh, and thanks for clarifying this documentation!

FWIW, it looks good to me. I let @dsagal make the final review.

help/en/docs/install/oidc.md
jordigh: Oops, thanks, fixed that.

help/en/docs/install/oidc.md
jordigh: How about extra newlines instead?

dsagal (approved):
This looks cleaner and clearer, a definite improvement. One thing I noticed is that `"openid email profile"` is already the default value for `GRIST_OIDC_IDP_SCOPES` -- is it still helpful to include it in the recommended configuration in these examples? Your call.

jordigh:
@dsagal Well, now I wonder why is it a config option at all. Grist already knows the scopes it needs, so why even make it configurable?

dsagal:
Good question, I don't know, only have a vague guess. @fflorent , do you know the answer to @jordigh's question?

fflorent:
Indeed, good question @jordigh!

Here are some explanations I have in mind:
 - The administrator may not want Grist to fetch some profile information (like the user's name), after all, Grist only needs the email address (so technically the only scopes required is `openid email` if I am right);
 - There are some information that we may want to let the administrator retrieve in the future, I think of enhancement in the user table to store [additional values](https://github.com/gristlabs/grist-core/issues/752) and use them in access rules;

These explanations may not be highly convincing (especially my second point). We may have introduced this env variable when needed or requested by the community.

jordigh:
@fflorent Okay, I thought Grist just wouldn't work if it didn't have all of the scopes, but if degraded performance is possible with fewer scopes than the default, that's good to know.

@dsagal Even if it's the default value, I think it's helpful to highlight in the examples that this value exists.

Given that, I'm merging this in.

PR #398, opened 2024-08-29
Move admonition to the top of the page

fflorent:
## Context

The admonition is thought to be displayed after the title of the page.

However, there are some glitch, where the admonition is displayed before the titles, like on the page about the [column types](https://support.getgrist.com/fr/col-types/), and sometimes even they do not appear at all, like the [functions reference](https://support.getgrist.com/fr/functions/) (don't take a look at the rendering, this is another PR I will submit ASAP).

## Proposed solution

I would say it is fine to display the admonition at the top, before the title. It simplifies the code.


dsagal (approved):

hexaltation (approved):
LGTM

Issue #404, opened 2024-08-31
[GitLocalize] Please invite me to the translation team


ricpol:
### GitLocalize account creation

- [X] My account is created, you can invite me

### Languages you would like to help translate

Others (please detail below)

### Other languages (optional)

Italian

fflorent:
Hello @ricpol!

I opened PR #405 to add the Italian language in the help support, I wait for a review and a merge from Grist Labs. I would then be able to add the Italian language in GitLocalize and you may then start translating.

BTW, I used some Python script I made quickly to ask GPT to translate every page. Would you like me to take time to cleanup this script, document it and make it available? 
It would require an Open AI api key. FWIW, it cost me 10$ to credit my api key and I consumed about 4$ to translate everything.

ricpol:
ok thanks I'll be super-busy with my day job for a month or so anyway, so I'll just wait... 
I think I'll translate by hand anyway... I have already translated all the Grist GUI, and I also maintain an [ongoing Italian translation](https://github.com/ricpol/pytutorial-it) of the official python tutorial, so... I'll be fine I think!

vviers:
Bravissimo !

PR #405, opened 2024-09-02
Add Italian language

fflorent:
In #404, a user requested to be added to GitLocalize to start translating in italian. 

This PR setup the folder `help/it` for that and adds the Italian option in translation template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/10-help-translating.yml
+++ b/.github/ISSUE_TEMPLATE/10-help-translating.yml
@@ -25,6 +25,7 @@ body:
       multiple: true
       options:
         - French
+        - Italian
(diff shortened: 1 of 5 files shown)


paulfitz (approved):
Thanks @fflorent (and @ricpol !)

ricpol:
Thanks guys! However, from the GiltLocalize side of it, it still looks like there is only the French translation available, like this: 

gitloc

Is there anything I should do first?

fflorent:
@ricpol Indeed. This should be better now. Could you confirm?

ricpol:
much better now, thanks! I'll get to work soon, promise!

PR #426, opened 2024-09-25
Two-Way References

nataliemisasi:
adding two-way-reference section to Reference columns article


help/en/docs/col-refs.md
```diff
@@ -68,6 +68,18 @@ the underlying table and insert the proper reference:
 *![Add reference value](images/column-ref-add-value.png)*
 {: .screenshot-half }
 
+## Creating a two-way Reference
+
+By default, new reference columns are created as a one-way reference, meaning that the reference is only shown in a single column in the table you have added it in. Grist also allows you to configure a reference column as a two-way reference, meaning that the reference will be shown in the column you initially added in the source table and as a reverse reference column in the referenced table.
```
jr-grist, line 73:
```suggestion
By default, new reference columns are created as one-way references, meaning that the references are only shown in a single column in the table you have added it in. Grist also allows you to configure a reference column with two-way references, meaning that the references will be shown in the column you initially added in the source table and in a reverse reference column in the referenced table.
```

help/en/docs/col-refs.md
```diff
@@ -68,6 +68,18 @@ the underlying table and insert the proper reference:
...
+
+For example, we can make the existing reference column, **Client**, a two-way reference. Under the 'Column' tab of the Creator Panel, select 'Add two-way reference'.
+
+*![two-way-ref](images/columns/two-way-ref.png)*
+
+This creates a reference column in the referenced table, *Clients*. Because there could be multiple projects for a single client, the **Projects** column is a [Reference List](col-types.md#reference-list-columns). An update made in one reference column will be reflected in the other.
```
jr-grist, line 79:
```suggestion
This creates a reverse reference column in the referenced table, *Clients*. Because there could be multiple projects for a single client, the **Projects** column is a [Reference List](col-types.md#reference-list-columns). An update made in a reference column will be reflected in the corresponding reverse reference column.
```

jr-grist (approved):
Added more clear language to distinguish between 'reference column' and 'reverse reference column'

nbush (changes requested):
This is a concise summary of the feature, and I think fine to release in its current state pending your review of my changes. I made some updates based on the language used in the UI to help keep things consistent (omitting "reverse", using "target" for example).

There are a few fringe elements that are still undocumented:
1. The "Record already assigned" warning you get when trying to add references to multiple records where that is not allowed (i.e. not a RefList).
2. The (somewhat unintuitive) behaviour when deleting a two-way reference (it also deletes a column)
3. The note that you can't use two-way references on Formula columns

I don't think these are critical for now, but are probably worth documenting at some point.

nataliemisasi:
reviewed nick's changes. they look good

nbush (approved):
Follow-up task created.

PR #431, opened 2024-10-10
Added ex. showing how to reorder conditional rules

jr-grist:
Added text and animation showing how to drag-and-drop conditional rules into a new order


help/en/docs/conditional-formatting.md
```diff
@@ -29,10 +29,13 @@ To add conditional formatting to rows, go to the `ROW STYLE` section of the [cre
 Order of Rules
 --------------
 
-Note that Grist applies the rules in order. Styles applied by later rules will override those applied by earlier rules.
+Note that Grist applies the rules in order. Styles applied by later rules will override those applied by earlier rules. To change the order of the rules, click and drag the appropriate rule to its new position using the drag handles to the left of each rule.
```
nbush, line 32: I know the animation shows it, but it's really disorienting for it to say "drag handles" here but them be invisible in the screenshot immediately after. I think we need something like:
> To change the order of the rules, click and drag the appropriate rule to its new position using the drag handles to the left of each rule. Note: These handles only appear once you hover over each rule.

That's a bit chunky, but I think it should be called out somewhere.

nbush (changes requested):
Small clarification requested.

help/en/docs/conditional-formatting.md
```diff
@@ -29,10 +29,13 @@ To add conditional formatting to rows, go to the `ROW STYLE` section of the [cre
 Order of Rules
 --------------
 
-Note that Grist applies the rules in order. Styles applied by later rules will override those applied by earlier rules.
```
nataliemisasi, line 32: What does this mean? Specify that the order is top to bottom

help/en/docs/conditional-formatting.md
nataliemisasi: You could also include a small screenshot right after this paragraph that includes the two vertical bars.
image

nataliemisasi (changes requested):

help/en/docs/conditional-formatting.md
jr-grist, line 32:
```suggestion
Note that Grist applies the rules in order from top to bottom. Styles applied by rules that appear later in the list will override those applied by rules earlier in the list. To change the order of the rules, hover over a rule to reveal a drag handle, then click and drag the drag handle to move the rule to its new position.
```

help/en/docs/conditional-formatting.md
jr-grist: See suggestion on Nick's comment below for resolution

help/en/docs/conditional-formatting.md
jr-grist: I think a text-based solution is preferable to including another screenshot, but I can go that route if this still isn't clear

jr-grist (commented):
Addressed comments re: clarifying list order, context for drag handles

help/en/docs/conditional-formatting.md
nbush: One last suggestion to remove the extra "drag": "hover over a rule to reveal a drag handle, then click and drag the drag handle" -> "hover over a rule to reveal a drag handle, then click and drag the handle"

Otherwise the above commit is good!

nbush (changes requested):
One more tiny change.

help/en/docs/conditional-formatting.md
nataliemisasi: I agree with Nick

nataliemisasi (changes requested):
See Nick's comment about "drag". Otherwise, looks good 👍

nataliemisasi (approved):
Looks good, thanks!

PR #432, opened 2024-10-10
Include extra info for webhook "Ready column" parameter usage

jmbenedetto:
When trying to use webhooks with "Ready column" option, I couldn't make it work. I asked help for the Grist team and they explained that:

"The second issue, with the "Ready" column not triggering, can be resolved by updating your "Event Types" to include "add" (in addition to "update"). The reason for this is because the “Ready” column actually determines whether the webhook logic sees the record as existing or not existing. So a change to the “Ready” column is seen as an “Add” event. Once “Ready” is true, then other changes to the same record are seen as “Update” events."

Thanks Natalie White for the explanation.

To help the next user, I'd like to include the core of this explanation on the webhook help page.

### DIFF
--- a/help/en/docs/webhooks.md
+++ b/help/en/docs/webhooks.md
@@ -46,9 +46,7 @@ the processing of the webhook. Not all fields are required.
   triggers when adding a row, it does not matter which columns are defined
   when the new row is added.
 * **Ready column** (optional): A boolean, or [Toggle](col-types.md#toggle-columns), column on the table that
-  determines if the row should trigger the webhook or not. When the
-  column becomes true, the corresponding row will trigger the webhook.
-  Rows become visibile to the webhook only when "Ready column" is first set to true. For this reason, we need both "add" and "update" event types selected for it to work.
+  determines if the row is visible to webhooks. While the toggle in the "Ready column" is false, changes to that row do not trigger the webhook. When the column becomes true, the row becomes visible to the webhook, and triggers the "add" event. While the column is true, further changes to the row trigger "update" events.
 * **URL**: The remote URL of the service that the webhook will notify
   of added or changed rows. When self-hosting, only external services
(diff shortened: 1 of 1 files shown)


dsagal:
What do you think of this revision, trying to make it even clearer:

> A boolean, or [Toggle](col-types.md#toggle-columns), column on the table that determines if the row is visible to webhooks. While the toggle in the "Ready column" is false, changes to that row do not trigger the webhook. When the column becomes true, the row becomes visible to the webhook, and triggers the "add" event. While the column is true, further changes to the row trigger "update" events.

jmbenedetto:
Pretty good!Thanks for spending time on this.Best regards JM Benedetto+55 11 99999-6764On 10 Oct 2024, at 16:36, Dmitry ***@***.***> wrote:﻿
What do you think of this revision, trying to make it even clearer:

A boolean, or Toggle, column on the table that determines if the row is visible to webhooks. While the toggle in the "Ready column" is false, changes to that row do not trigger the webhook. When the column becomes true, the row becomes visible to the webhook, and triggers the "add" event. While the column is true, further changes to the row trigger "update" events.

dsagal:
Hi @jmbenedetto -- if you don't mind making the change in this PR, this would be the easiest way to approve and merge it. Thanks again for taking the initiative!

jmbenedetto:
Sorry, Dimitri! Will do. It is my first time contributing to a project. JM Benedetto+55 11 99999-6764On 31 Oct 2024, at 02:14, Dmitry ***@***.***> wrote:﻿
Hi @jmbenedetto -- if you don't mind making the change in this PR, this would be the easiest way to approve and merge it. Thanks again for taking the initiative!

jmbenedetto:
Done.

dsagal (approved):
Thank you!

PR #441, opened 2024-11-05
Clarification on fields added in form widget

nataliemisasi:
clarifying fields added in forms widget are hidden by default in table widgets


nbush (approved):
👍

PR #447, opened 2024-11-23
marketplace: write instructions for transitioning

jordigh:
There are very few users who should ever need to do this, so the instructions are a little sparse on details. I believe this to be an acceptable compromise.

PR #448, opened 2024-11-25
self-hosted: document the importance of `GRIST_DEFAULT_EMAIL`

jordigh:
I need to reference this value from [another PR](https://github.com/gristlabs/grist-help/pull/447/files)

### DIFF
--- a/help/en/docs/self-managed.md
+++ b/help/en/docs/self-managed.md
@@ -61,6 +61,7 @@ To try Grist out using Docker, make an empty directory for Grist to store materi
 docker run -p 8484:8484 \
   -v ~/grist:/persist \
   -e GRIST_SESSION_SECRET=invent-a-secret-here \
+  -e GRIST_DEFAULT_EMAIL=your-email@example.com \
   -it gristlabs/grist
 ```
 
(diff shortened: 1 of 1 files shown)


paulfitz:
An extra gotcha that often trips people up is ownership of the team site created on first run if `GRIST_SINGLE_ORG`  is set
https://github.com/gristlabs/grist-core/blob/1f984ab5a7338430ab98cba42cb0d7b0daa229bc/stubs/app/server/server.ts#L63-L90
The owner will be `GRIST_DEFAULT_EMAIL` so things get very confusing if that email is changed but the team site is kept (and is now inaccessible!)

jordigh:
I added a comment about this. Does that help?

help/en/docs/self-managed.md
```diff
@@ -237,6 +254,14 @@ The name of the team should use only the lower-case characters a-z, the digits
 [Custom styling](self-managed.md#how-do-i-customize-styling) to hide any UI elements
 you don't need.
 
+When [using
+`GRIST_SINGLE_ORG`](self-managed.md#how-do-i-set-up-a-team), the
+single org will also be created initially and owned by the user
+configured by `GRIST_DEFAULT_EMAIL`. If you change the value of
+`GRIST_DEFAULT_EMAIL`, [the
+administrator](self-managed.md#what-is-the-administrative-account) may
+temporarily lose access to the team site.
```
paulfitz, line 263: This could be a little more precise.

If `GRIST_SINGLE_ORG` is set, then on start-up Grist will create the specified team site if isn't already present. When creating it, it will make the current administrator an owner of it. Before changing the administrator, it is a good idea to use [team sharing](https://support.getgrist.com/team-sharing/#team-sharing) options to add new owners to the team site, since apart from initial creation, administrators have no automatic right to team site membership.

paulfitz (commented):
Thanks Jordi! One more note.

help/en/docs/self-managed.md
jordigh: I added your note. Does that look better?

paulfitz (approved):
Thanks @jordigh

PR #453, opened 2024-12-11
adding a link from Exports to self-managed documentatioon

nataliemisasi:
Adding a link from Exports to Self-managed documentation per customer suggestion


nataliemisasi:
Decision was that this may have been a one-off issue and we do not feel like changes are necessary. If it comes up again, we can improve self-managed to show up in search for specific terms

PR #467, opened 2025-01-09
Externalize scim endpoints parameters

berhalak:
Moving parameters from `scim` endpoints to shared `components` section. It is not needed now, but may be needed in the future. Additionally added a new script `./api/parameters.js` that helps with the process.

Related:
Read more in https://github.com/gristlabs/grist-help/pull/293


api/scim/users.yml
```diff
@@ -417,3 +399,28 @@ components:
             description: The unique identifier of the user.
             example: "1"
       - $ref: '#/components/schemas/UserInRequest'
+  parameters:
```
georgegevoian, line 402: I see there are already some keys in `params`. Any reason why we shouldn't put these there too?

PR #475, opened 2025-02-17
Add snapshot env variables and recommendation about lifecycle rules

georgegevoian:
This documents the variables added in [this PR](https://github.com/gristlabs/grist-core/pull/505), and adds a recommendation about maintaining snapshot metadata with lifecycle rules.

### DIFF
--- a/help/en/docs/install/cloud-storage.md
+++ b/help/en/docs/install/cloud-storage.md
@@ -78,11 +78,17 @@ snapshot metadata are pruned regularly.
 
 You can control the frequency of snapshots with the following environment variables:
 
-  * `GRIST_SNAPSHOT_TIME_CAP` - JSON string formatted as: `{"hour": 25, "day": 32, "isoWeek": 12, "month": 96, "year": 1000}`:
-    * (Required) `hour` - The number of distinct hours to keep a snapshot for. (Default: 25)
-    * (Required) `day` - The number of distinct days to keep a snapshot for. (Default: 32)
-    * (Required) `isoWeek` - The number of distinct weeks to keep a snapshot for. (Default: 12)
-    * (Required) `month` - The number of distinct months to keep a snapshot for. (Default: 96)
-    * (Required) `year` - The number of distinct years to keep a snapshot for. (Default: 1000)
+  * `GRIST_SNAPSHOT_TIME_CAP` - JSON string specifying the maximum number of backups to keep
+  for each time period (e.g. hourly, daily, weekly). For example, a value of
+  `{"hour": 24, "day": 30, "isoWeek": 52, "month": 24, "year": 5}` will keep the most recent
+  backup for every hour (for the last 24 hours), day (for the last 30 days), etc.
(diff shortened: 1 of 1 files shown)


help/en/docs/install/cloud-storage.md
```diff
@@ -69,5 +69,20 @@ without any preexisting Grist documents.
...
+You can control the frequency of snapshots with the following environment variables:
+
+  * `GRIST_SNAPSHOT_TIME_CAP` - JSON string formatted as: `{"hour": 25, "day": 32, "isoWeek": 12, "month": 96, "year": 1000}`:
+    * (Required) `hour` - The number of distinct hours to keep a snapshot for. (Default: 25)
+    * (Required) `day` - The number of distinct days to keep a snapshot for. (Default: 32)
+    * (Required) `isoWeek` - The number of distinct weeks to keep a snapshot for. (Default: 12)
+    * (Required) `month` - The number of distinct months to keep a snapshot for. (Default: 96)
+    * (Required) `year` - The number of distinct years to keep a snapshot for. (Default: 1000)
+  * `GRIST_SNAPSHOT_KEEP` - Maximum number of recent snapshots to keep, regardless of
+  `GRIST_SNAPSHOT_TIME_CAP`. (Default: 5)
```
Spoffy, line 88: I don't understand how these interact.

E.g - If I set "hour" to 12, "day" to 3 and "month" to 4, when do my snapshots get deleted?

Spoffy (changes requested):

help/en/docs/install/cloud-storage.md
georgegevoian: I'll update the description to clarify better, but for your example, we'd keep at most one snapshot each of the last 12 hours, 3 days, and 4 months. The snapshot kept in each period is the most recent one that occurred, so the ones that get deleted are previous ones taken within the same time period (e.g. hour, day).

`GRIST_SNAPSHOT_KEEP` works separately from the bucketing scheme above - it'll always keep the N most recent snapshots, even if they were taken minutes apart.

paulfitz: Some backstory (good to add to commit message @georgegevoian, by default that is now the top comment) https://github.com/gristlabs/grist-core/pull/505

To add to what @georgegevoian is saying: here is the default logic for keeping a snapshot -

```
/**
 * Calculate which snapshots to keep.  Expects most recent snapshots to be first.
 * We keep:
 *   - The five most recent versions (including the current version)
 *   - The most recent version in every hour, for up to 25 distinct hours
 *   - The most recent version in every day, for up to 32 distinct days
 *   - The most recent version in every week, for up to 12 distinct weeks
 *   - The most recent version in every month, for up to 96 distinct months
 *   - The most recent version in every year, for up to 1000 distinct years
 *   - Anything with a label, for up to 32 days before the current version.
 * Calculations done in UTC, Gregorian calendar, ISO weeks (week starts with Monday).
 */
 ```

This variable allows tweaking that default.

Spoffy (approved):

PR #476, opened 2025-02-17
Adds external storage end-user docs

Spoffy:
Adds documentation explaining external attachments to document owners.

This primarily covers the consequences of using external attachments, and how to enable them on a document.

This doesn't contain any documentation for admins on how to enable external attachments on their Grist instance.


help/en/docs/document-settings.md
```diff
@@ -43,3 +43,35 @@ The API section provides tools for integrating your document with external servi
...
+- Allow large files to be used as attachments, without impacting performance.
+- Allow significantly more files to be attached, without impacting performance.
+- Speed up saving and restoring document snapshots by reducing their size.
+
+However, there are additional limitations placed on documents that use external attachments:
+
+- Any downloaded documents will have broken attachments. These will remain broken even if the document is re-uploaded.
+- Duplicating a document will be disabled. However, you will still be able to work on a copy of the document, and save the changes back to the original document.
+
+Therefore, it's advised you only use external attachments if your document is having problems due to attachments.
```
nataliemisasi, line 65: Explain what kind of problems. The one I'm aware of would be document slowness. Are there any others? Would be good to be specific here so users will think "oh, my document does seem slow - maybe this could help!" It also is worth noting that the limit of attachments + data is 1GB and they can check their usage under Raw Data. That could be useful in a blue info box
Possible suggestion (Feel free to reword further):
```suggestion
Therefore, it's advised you only use external attachments if your document is experiencing slowness due to attachments.
```
Also add a blue box with the following language: "Attachments plus data in a single document are limited to 1GB on all plans. You can check your document's usage under [Raw Data](raw-data.md)."
Title could be Attachment Limits or something like that?

help/en/docs/document-settings.md
```diff
@@ -43,3 +43,35 @@ The API section provides tools for integrating your document with external servi
...
+
+External attachments can be enabled from the Document Settings page. If external attachments are not available on your Grist installation, the dropdown will be disabled.
+
+<span class="screenshot-large">*![attachment-storage-setting](images/document-settings/attachment-storage-setting.png)*</span>
+
+A transfer process can be started to move attachments between external and internal storage. This transfer may be paused due to document inactivity. If a transfer is paused due to document inactivity, you can safely resume it anytime from the document settings page.
+
+<span class="screenshot-large">*![attachment-storage-transfer](images/document-settings/attachment-storage-transfer.png)*</span>
+
+The transfer process may take some time, depending on the size and number of attachments.
```
nataliemisasi, line 77:
```suggestion
The transfer process may take some time, depending on the size and number of attachments. 

<span class="screenshot-large">*![attachment-storage-transfer](images/document-settings/attachment-storage-transfer.png)*</span>
```

help/en/docs/document-settings.md
Spoffy: Snapshot upload / restore slowness is the one I know about, I'm not sure what other slownesses might exist. Copy time and working copy time perhaps? I'll need to put a test document together to check.

Spoffy: For the blue box - do we want to add that now, or later given:
* It's not available in the SaaS yet (and there's no limit on self-managed?)
* The 'Raw data' usage tracker doesn't work for external attachments (right now)

nataliemisasi: > Snapshot upload / restore slowness is the one I know about, I'm not sure what other slownesses might exist. Copy time and working copy time perhaps? I'll need to put a test document together to check.

Ok then I think we can stick with my suggested text "...if your document is experiencing slowness due to attachments". Then, if they search 'slowness', they'll find this and it gives them a place to begin with troubleshooting 👌 

> For the blue box - do we want to add that now, or later given:
> 
> * It's not available in the SaaS yet (and there's no limit on self-managed?)
> * The 'Raw data' usage tracker doesn't work for external attachments (right now)

Add later: I'll create a task to add it when it's available in SaaS (and to remove the yellow box at the top saying this is only available for self-hosted). When I add it, I can use the title 'Internal Attachment Limits' so it's specific to those stored within the document 👍

Spoffy: Perfect, thanks! I'll make those changes shortly :)

nbush:
Re: [this question](https://discord.com/channels/1176642613022044301/1176642613663760456/1341546602342518877) on Discord, is there a way to incorporate at least Vincent's response so that people have an idea of how to take advantage of this feature? Or is there a reason we're not documenting it at the moment?

nataliemisasi (approved):
Looks good!

Spoffy:
> Re: [this question](https://discord.com/channels/1176642613022044301/1176642613663760456/1341546602342518877) on Discord, is there a way to incorporate at least Vincent's response so that people have an idea of how to take advantage of this feature? Or is there a reason we're not documenting it at the moment?

I'd say this is a separate bit of documentation for self-hosters, that we need to nail down in another PR :)

PR #477, opened 2025-02-20
snapshots: rewrite the various commands into a single Compose setup

jordigh:
Running a series of Docker commands plus initialisation commands is a bit daunting. There are several steps to take, each of which could be subtly wrong. It seems very error-prone to me.

This example condenses all of those steps into just two files and a single command to run: `docker compose up`. Hopefully this will be a better explanation and easier to follow.

Thanks to @Spoffy for the initial Compose setup that I worked from.

### DIFF
--- a/help/en/docs/self-managed.md
+++ b/help/en/docs/self-managed.md
@@ -724,50 +724,105 @@ Grist's cloud storage feature allows automatic syncing of Grist
 documents and document versions to an S3-compatible bucket
 (available for all Grist versions) or to Azure storage (in Enterprise Grist).
 
-Here is an example of running Grist locally, with snapshots stored
-in a test MinIO instance:
+!!! warning "Warning"
+
+    **It is entirely possible to permanently lose documents when
+    configuring snapshots. We strongly urge users to back up their
+    documents (`.grist` files) data before attempting to follow these
+    instructions.**
+
+Below is an example Docker Compose setup for running Grist locally,
+with snapshots stored in a test MinIO instance. First, create a `.env`
+file with the following example configuration:
 
 ```sh
-# Make a network
-docker network create grist
+MINIO_PASSWORD=gristadmin
+PERSIST_DIR=/tmp/persist
 
-# Start Redis in our network (recommended for snapshots)
-docker run --rm --network grist --name redis redis
-
-# Start MinIO in our network
-docker run --rm --network grist --name minio \
-  -v /tmp/minio:/data \
-  -p 9000:9000 -p 9001:9001 \
-  -e MINIO_ROOT_USER=grist -e MINIO_ROOT_PASSWORD=admingrist \
-   -it minio/minio server /data -console-address ":9001"
-
-# Visit http://localhost:9000 and set up a bucket called grist-docs.
-# Make sure to enable versioning on the bucket.
-
-# Hook Grist up to Redis and MinIO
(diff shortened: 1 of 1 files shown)


paulfitz:
Hmm this feels like the wrong direction for documentation? Setting up snapshots is currently hard, you have to know a bunch of stuff. The docker compose template could help if you are just trying to do some local testing, but then you'll need to actually understand all the bits and pieces when you go to do something real. I feel because there are more lines here, people are more likely to copy and paste than they are to understand something shorter that was never intended to actually be literally run.

The ideal is an admin section that helps you get going step by step, and offers diagnostics. Like is there a bucket attached. Does it have versioning set up. etc.

Linking to a docker compose example would be fine.

I might well be wrong on this, so suggest soliciting a couple points of view.

paulfitz:
Btw some really good information to add in an update to the snapshot docs is "IF YOU HAVE VALUABLE DATA ALREADY, REALLY TRULY MAKE BACKUPS OF ALL YOUR DOCS BECAUSE YOU ARE VERY LIKELY TO LOSE THEM AT LEAST ONCE TRYING THIS." It can be a traumatic transition right now, like migrating to Postgres.

jordigh:
> some really good information to add in an update to the snapshot docs

I've added a warning to this effect.

help/en/docs/self-managed.md
```diff
@@ -724,50 +724,105 @@ Grist's cloud storage feature allows automatic syncing of Grist
...
+
+      GRIST_DOCS_MINIO_USE_SSL: 0
+      GRIST_DOCS_MINIO_PORT: 9000
+      REDIS_URL: redis://redis
+
+      # Add any additional Grist configuration here
+      #
+      # ...
+    ports:
+      - 8484:8448
```
paulfitz, line 775: Why 8448 btw?

jordigh: Oops. I swear I actually ran this example before publishing it. Not sure how this happened. Fixed!

Spoffy (changes requested):
I think this is an improvement over the original docs, but might benefit from a little more explanation before we leap into the example. 

Something along these lines:
> Grist's cloud storage is automatically enabled when Grist is configured with access to an external store. This is done by setting specific environment variables for the desired store (e.g. MinIO).
>
> In addition to storage, connecting Grist to a Redis instance is beneficial to snapshot storage for <reasons>.
>
> The example compose file below demonstrates a minimum setup to use Grist with Minio as a snapshot store. It shows the environment variables that need to be set, and starts up an example Grist instance with corresponding MinIO and Redis services. 
> 
> Additionally, it configures MinIO for snapshot storage. Crucially it:
> - Creates a new bucket for Grist to use.
> - Enables versioning on the bucket.
> - Enables anonymous access.

Then proceed to the example. So we're explaining:
- The main settings Grist needs
- How MinIO needs to be set up.

As it's implicit in the example right now, but would be nice to be explicit up front :)

*Edit: Reading Paul's comments, I think I'm agreeing with him? The most benefit I can see is from explaining the requirements / steps to a user, rather than handing over an example. Although I also don't see an issue with keeping / linking the compose file - I think it's clearer than the actual docker commands.*

*We could split the compose file up  into each service, and explain it perhaps?*

paulfitz:
Closing, stale.

PR #478, opened 2025-02-21
builder-edition: add a link to the technical README

jordigh:
So that people are more aware that more documentation exists.


paulfitz (approved):
Thanks @jordigh !

PR #479, opened 2025-02-26
update integrations

nataliemisasi:
add albato to list of integrator servicers


nataliemisasi:
discussion with Dmitry & Nick led to decision to not add to list of integrators until they have users using the Grist integration. closing this PR

PR #488, opened 2025-03-17
Adds API docs for "POST /docs"

Spoffy:
Adds API documentation for POSTing to the "/docs" endpoint.

The `/docs` endpoint has four behaviours:
1. Copy an existing document
2. Upload a new document from a file
3. Create a new document in a specific workspace
4. Create a new unsaved document

Each documented using its own schema.


Spoffy:
This overlaps with this endpoint: `/workspaces/{workspaceId}/docs`, which creates a new, saved doc in the given workspace.

This means it looks odd in the API documentation, having two "Create new doc" endpoints:
image

My thinking is we avoid documenting use cases 3. and 4. (creating [un]saved documents), and document this purely as a Copy/Upload endpoint.

Any thoughts would be appreciated!

Spoffy:
On hold until the underlying API can be de-overloaded

PR #493, opened 2025-03-31
Add a note that Grist 1.5.0 disables Postgres JIT compilation.

jordigh:


help/en/docs/self-managed.md
```diff
@@ -687,9 +687,9 @@ and set the following variables:
 * TYPEORM_PORT - set to port number of database if not the default for PostgreSQL
 
 Grist is known to work with PostgreSQL from versions 10 through 16.
-Recent versions, however, have enabled by default a JIT compiler that
+Version 12 and onwards, however, has enabled by default a JIT compiler that
```
paulfitz, line 690: I think I'd still expect `have` because `Version 12 and onwards` is multiple versions? Depends how you read it I guess, your call then.

paulfitz (approved):
Thanks for following up @jordigh !

help/en/docs/self-managed.md
jordigh: Thanks, I've rephrased this.

paulfitz (approved):

PR #498, opened 2025-04-08
linking admin panel

nataliemisasi:
adding links to admin controls from self-managed article


dsagal (commented):
Mmm, there is Admin Panel -- which apparently we don't have documented? It's mentioned for example here https://support.getgrist.com/newsletters/2024-05/#self-hosting-admin-console-improvements, and is available to all self-hosters. Open-source ones could use it to upgrade to Enterprise, for example, since that upgrade is one of Admin Panel's featuers.

Then there are Admin Controls. That's only enterprise, and rather than configuring the installation, it allows viewing and managing sites/workspaces/docs/users.

But linking Admin Panel to the article on Admin Controls would only make things more confusing, I think.

nataliemisasi:
CONFIRMED:
Admin Panel - available to Self-Hosters

Admin Controls - only available to Enterprise users

New article - Admin Panel with two short sections, settings and admin controls. link to admin controls article

help/en/docs/self-managed.md
```diff
@@ -830,6 +832,10 @@ Utilities such as [Watchtower](https://containrrr.dev/watchtower/) can
 keep your version of Grist up to date for you. We can also install and maintain self-hosted 
 Grist installations for Enterprise users. [Contact us](https://www.getgrist.com/contact/) for more information.
 
+### How do I remove a user from my site? {: .tag-ee }
```
dsagal, line 835: Suggest changing to "How do I completely remove a user from my instance"? (To avoid confusion with removing a user from team site)

help/en/docs/admin-panel.md
```diff
@@ -0,0 +1,27 @@
...
+
+*![Admin Panel in tools](images/admin-controls/admin-panel-in-tools.png)*
+{: .screenshot-half }
+
+*![Admin Panel in user menu](images/admin-controls/admin-panel-in-user-menu.png)*
+{: .screenshot-half }
+
+## Settings
+
+The Installation page is where you can configure installation settings including [Telemetry](telemetry.md), Security Settings and Versioning. If you are on an Enterprise plan, you'll also have access to [Audit Logs](install/audit-log-overview.md).
```
dsagal, line 19: Suggest "you'll also be able to configure [Audit Logs]"

dsagal (approved):
Thanks!

PR #500, opened 2025-04-19
[FIX] Remove tab character after property in users.yml

QazCetelic:
This tab character causes problems parsing the file when using [openapi-generator](https://github.com/OpenAPITools/openapi-generator).


fflorent (approved):
LGTM, thanks!

Issue #501, opened 2025-04-20
OpenAPI spec issues


QazCetelic:
- Path `/docs/{docId}/states/remove` doesn't specify the possible responses
- `users.yml` file contains an invalid tab character on line 401
- `WebhookProperties` component is invalid

PR #502, opened 2025-04-20
fix(openapi-spec): Fixed some of API specification issues

QazCetelic:
This fixes the other 2 issues mentioned in #501

There is already a PR (#500) for the tab character issue that is mentioned in #501 already


paulfitz:
Sorry this PR got overlooked! If you're still available, could you rebase to resolve conflicts?

PR #512, opened 2025-05-13
api docs: fix sort options

iexos:
I couldn't get `choiceOrder` to work, so found its actually `orderByChoice`. Found another error in the same place

correct values in the source [here](https://github.com/gristlabs/grist-core/blob/7a09adcf0b1838667bba4a585403a95169df38c4/app/client/ui/SortConfig.ts#L153)

### DIFF
--- a/api/grist.yml
+++ b/api/grist.yml
@@ -2299,7 +2299,7 @@ components:
       name: sort
       schema:
         type: string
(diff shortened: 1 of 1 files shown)


paulfitz (approved):
Thank you!

PR #513, opened 2025-05-19
Fix typo in cloud-storage.md

vviers:
My understanding of this variable is that it is a minimum number of snapshots to keep, regardless of the rest


georgegevoian (approved):
Thanks @vviers.

PR #515, opened 2025-05-27
Adds copy endpoint docs

Spoffy:
Adds API documentation for the copy endpoint


api/grist.yml
```diff
@@ -330,6 +330,42 @@ paths:
...
+      parameters:
+        - $ref: '#/components/parameters/docIdPathParam'
+      requestBody:
+        description: "the target workspace"
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - workspace
```
dsagal, line 348: I think this list should rather include `workspaceId` and `documentName`, if I understand correctly that those are the required properties.

api/grist.yml
```diff
@@ -330,6 +330,42 @@ paths:
...
+                  example: 597
+                documentName:
+                  type: string
+                  example: "Example Doc"
+                asTemplate:
+                  type: boolean
+                  description: "Remove all data and history but keep the structure to use the document as a template"
+      responses:
+        200:
+          description: "Success"
```
dsagal, line 361: How about using the same description as for `createDoc` (e.g. "description" would be "The document id", better example, etc.

Spoffy: Great idea, updated this and the description of the schema to match createDoc.

dsagal (approved):
Thanks!

PR #527, opened 2025-07-02
assistant-updates

nataliemisasi:
Updating assistant cans and cants

### DIFF
--- a/help/en/docs/assistant.md
+++ b/help/en/docs/assistant.md
@@ -11,8 +11,8 @@ When asking the Assistant a question, there’s no need to specify column IDs or
 
 The Assistant:
 
-* **Can** answer questions about your data, create or modify records, and help with formulas. It also knows which page you're viewing.
-* **Can't** create or modify tables, columns, views, or access rules. It also doesn't know if you've selected something.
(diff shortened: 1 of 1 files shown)


help/en/docs/assistant.md
```diff
@@ -11,8 +11,8 @@ When asking the Assistant a question, there’s no need to specify column IDs or
 
 The Assistant:
 
-* **Can** answer questions about your data, create or modify records, and help with formulas. It also knows which page you're viewing.
-* **Can't** create or modify tables, columns, views, or access rules. It also doesn't know if you've selected something.
+* **Can** answer questions about your data, create or modify tables, columns, or records, and help with formulas. It also knows which page you're viewing.
+* **Can't** create or modify pages, widgets, or access rules. It also doesn't know if you've selected something.
```
nbush, line 15: I tested on staging (because I'm out of credits otherwise), but the assistant can create pages when it's asked to make new tables. 

This is accurate in most cases though, as it gives you a "permissions" error when you ask to create a random page. I think we're good to keep it like this for now.

nbush (approved):

PR #528, opened 2025-07-02
External-attachments-limits

nataliemisasi:
Adding blue note box to External Attachments section stating internal attachment limits


nbush (approved):
👍

PR #530, opened 2025-07-11
adding external/internal context

nbush:
and fixing links

### DIFF
--- a/help/en/docs/limits.md
+++ b/help/en/docs/limits.md
@@ -49,7 +49,7 @@ For memory and performance reasons, there's a recommended data size limit of 20M
 
 ## Attachments
 
-Attachments are counted separately from data. Internal attachments (i.e. default [attachment columns](col-types.md#attachment-columns)) are limited to **1GB** per document in all cases. For [external attachments](document-settings.md#external-attachments), there is a per document limit as well as a total attachment limit across all documents within an organization. Below are the limits for each Grist plan:
+Attachments are counted separately from data. Internal attachments (i.e. default [attachment columns](col-types.md#attachment-columns)) are limited to **1GB** per document in all cases. For [external attachments](document-settings.md#external-attachments), there is a per document limit that's higher on paid plans. There is also a total attachment limit across all documents within an organization. Below are the limits for each Grist plan:
 
 | Plan | Document Limit |  Organization Limit  |
 | --- | --- | --- |


help/en/docs/limits.md
```diff
@@ -43,21 +43,21 @@ Documents are also subject to data size limits, as described below.
...
-| Plan | Document Limit |  Org Limit  |
+| Plan | Document Limit |  Organization Limit  |
 | --- | --- | --- |
 | Free Personal | 1GB | 50GB |
 | Free Team | 1GB | 50GB |
 | Pro | 3GB | 100GB |
 | Business | 20GB | 200GB |
-| Enterprise | 100GB | 1TB (installation limit) |
+| Enterprise | Custom | Custom |
 | AppSumo |  | 100GB |
```
dsagal, line 61: Let's remove the "AppSumo" line. It was in internal docs because we have to deal with it still, but it's nothing we have to document publicly. The limits should satisfy those of "Pro" plan, which they do, and we should list any different ones. (We don't mention AppSumo anywhere else in documentation either, I don't think.)

dsagal (commented):
Sorry, I am a little confused about the current state of limits for internal attachments. @paulfitz, don't we still enforce a 1GB limit on internal attachments? I think it's important to do that because our SaaS doc-workers' disk is limited, but I no longer see any enforcement of a lower limit when attachments are internal.

paulfitz:
> Sorry, I am a little confused about the current state of limits for internal attachments. @paulfitz, don't we still enforce a 1GB limit on internal attachments? I think it's important to do that because our SaaS doc-workers' disk is limited, but I no longer see any enforcement of a lower limit when attachments are internal.

@nbush after talking with @dsagal about this, can you document a 1GB limit on internal attachments on our SaaS? I've added a task to enforce it at the level of adding new attachments. Users will already see a nudge to use external attachments long before they reach this limit.

nbush:
> @nbush after talking with @dsagal about this, can you document a 1GB limit on internal attachments on our SaaS? I've added a task to enforce it at the level of adding new attachments. Users will already see a nudge to use external attachments long before they reach this limit.

The current text has "Attachment limits apply to both internal and [external attachments](https://deploy-preview-530--grist-help-preview.netlify.app/document-settings/#external-attachments)." to go along with the 1GB limit in the table for free accounts. Are you suggesting we split internal/external limits, or is something unclear?

help/en/docs/limits.md
```diff
@@ -43,22 +43,21 @@ Documents are also subject to data size limits, as described below.
...
+For memory and performance reasons, there's a recommended data size limit of 20MB. Documents beyond 20MB may slow down or run into memory limits depending on their complexity. As an example, a document with 100,000 rows and 24 numeric columns would reach this recommended limit. To help optimize formulas on large documents, you can use the built-in [formula timer](formula-timer.md).
 
 ## Attachments
 
-Attachments are counted separately from data. Each plan includes an attachment limit per document, as well as a total attachment limit across all documents within the organization.
+Attachments are counted separately from data. Internal attachments (i.e. default [attachment columns](col-types.md#attachment-columns)) are limited to **1GB** per document in all cases. For [external attachments](document-settings.md#external-attachments), there is a per document limit as well as a total attachment limit across all documents within an organization. Below are the limits for each Grist plan:
```
dsagal, line 52: Suggestion:
```suggestion
Attachments are counted separately from data. Internal attachments (i.e. default [attachment columns](col-types.md#attachment-columns)) are limited to **1GB** per document in all cases. For [external attachments](document-settings.md#external-attachments), there is a per document limit that's higher on paid plans. There is also a total attachment limit across all documents within an organization. Below are the limits for each Grist plan:
```

Issue #546, opened 2025-09-05
Add link new R client libraries available


jengelaere:
Hello, 

As asked on Mon, Jul 28, 2025 with a mail to support@getgrist.com, would it be possible to reference our new client library for R (https://github.com/spyrales/gristapi) just after JS and Python ones in page https://support.getgrist.com/rest-api/#usage ?

I can propose a MR to add the line :
` * [R client library](https://spyrales.github.io/gristapi/)`
just here : https://github.com/gristlabs/grist-help/blob/fc5bd12570d9acfb974b303bb9182429b210dc78/help/en/docs/rest-api.md?plain=1#L83
if it's OK and helps.

Thanks for this great tool.

paulfitz:
Thanks @jengelaere. We have been updating that list recently, and I know your library was discussed and slated for addition but then slipped through the cracks? A merge request would be much appreciated 🙏

jengelaere:
Thanks @paulfitz 🙏 
it's done here : 
[Update rest-api.md - Add reference to a client library for R #547](https://github.com/gristlabs/grist-help/pull/547)

PR #547, opened 2025-09-05
Update rest-api.md - Add reference to a client library for R

jengelaere:
related #546

### DIFF
--- a/help/en/docs/rest-api.md
+++ b/help/en/docs/rest-api.md
@@ -80,6 +80,7 @@ There are two client libraries developed and maintained by Grist Labs:
 
  * [JavaScript/TypeScript client library](https://www.npmjs.com/package/grist-api)
  * [Python client library](https://pypi.org/project/grist-api/)
+ * [R client library](https://spyrales.github.io/gristapi/)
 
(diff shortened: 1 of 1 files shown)


help/en/docs/rest-api.md
```diff
@@ -80,6 +80,7 @@ There are two client libraries developed and maintained by Grist Labs:
 
  * [JavaScript/TypeScript client library](https://www.npmjs.com/package/grist-api)
  * [Python client library](https://pypi.org/project/grist-api/)
+ * [R client library](https://spyrales.github.io/gristapi/)
```
paulfitz, line 83: This list is for clients that Grist Labs wrote and will do support for. Could you add to the list below instead? Just so we don't get support requests for something we don't know much about.

paulfitz (commented):
Thanks @jengelaere ! One comment.

help/en/docs/rest-api.md
jengelaere: 😳 sorry for that, I've not read enought, 
new proposition with commit 7852b7913b935c8fa5d63ea1b5bec40ebea409e9

PR #560, opened 2025-10-01
add-referenced-column-updates

nataliemisasi:
updating 'adding referenced columns' options to include lookups under 'add column' menu


help/en/docs/images/column-ref-other-columns2.png
nbush, line 1: Apologies if I'm wrong, but I found this confusing because it shows the state _after_ the Contact column has been added, unlike the image above.

nbush (approved):

PR #567, opened 2025-10-22
Expand section on Configuring Fields to explain the new options 'Hidden field' and 'Accept value from URL'

dsagal:


dsagal:
Preview of the feature here: https://grist-dsagal-grist-form-url-fields.fly.dev/ucNDxbHRUnLe/Sample-Form/p/3

PR #570, opened 2025-10-26
Document the delete endpoint

jordigh:


jordigh:
Thanks. Looks like I also had to build the API docs and commit the build to the repo.

PR #571, opened 2025-10-26
Document that notifications also require Redis

jordigh:


paulfitz (commented):
Thanks @jordigh. This is useful. But it would be even more helpful to mention the requirement in a section a user might go to when wondering about how to configure notifications, or whether they are missing a step in configuring notifications. They might end up here:
https://support.getgrist.com/self-managed/#how-do-i-set-up-email-notifications

jordigh:
I've added another description about the need for Redis.

PR #589, opened 2025-12-28
Fix images in <center> tags; fix most anchor links that build warns about

dsagal:
- `<center>` tag now needs to be `<center markdown>` to process image markup inside.
- Also fixed various anchor links in /en/ docs.


dsagal:
@manuhabitela , a bunch of these affect /fr/ docs too. I fixed the `<center>` tags, but for other warnings, I noticed that the build gives a ton more warnings about /fr/ files, I think mainly due to changes since they were translated. I decided not to attempt fix-ups, but the warnings can be seen by running `./docs.py build-lang fr`.

PR #596, opened 2026-01-21
Fix french version issues

manuhabitela:
- fix lots of same-page anchors that were in English while they had to be translated
- improve the language switcher in the header
- make some images work in the French translation. My logic was to… just copy the missing files in the French build. Making each build independant. As it was already done for the /images directory. Not sure this logic is good to follow for everything?

_this is still a draft while I take the time to verify everything._


manuhabitela:
Thanks @dsagal :)

PR #599, opened 2026-01-28
access rules style tweaks

nbush:


dsagal:
Don't worry about netlify failure, I think it's getting fixed by George in his other grist-help PR.

PR #614, opened 2026-02-27
Adds docs for self-hosted Airtable setup

Spoffy:
Notable changes:
- Re-named "Integrations" to "Building integrations" for clarity. That section is entirely about writing / creating external services to interact with Grist.
- Added "Self-hosted / Configuring integrations" to hold Airtable and any future integration setup we might want. Google Drive is noticeably lacking from the current docs, as far as I can see.
- Added documentation for self-hosted Airtable setup. 
  - Note: I used a table here for environment variables, as I felt it was more legible than the way they're expressed elsewhere in the docs (e.g. Authentication / SAML). The other sections describing environment variables may also benefit from being migrated to a table. 
  
  
image

### DIFF
--- a/help/en/docs/install/integrations/airtable.md
+++ b/help/en/docs/install/integrations/airtable.md
@@ -2,7 +2,7 @@
 title: Airtable
 ---
 
-# Airtable Integration
+# Airtable integration
 
 ## Importing from Airtable
 


Spoffy:
Agree on the table feeling weird on mobile - might want a CSS change there with the rendering? It seems to prioritise column spacing oddly. 

Bullets would be better on mobile, but feel really hard to read. There's probably some compromise in here that makes sense.

PR #644, opened 2026-04-23
Documents the ?expand=column syntax for tables endpoint

Spoffy:
The /tables endpoint recently had the ?expand=column parameter added, to allow listing all of the column info nested within the table metadata.

This adds documentation for that with an example.


paulfitz (approved):
Thanks @Spoffy !

PR #649, opened 2026-05-13
Document first-run boot login and Quick setup wizard

paulfitz:
A new help page that walks self-hosting admins through what they see the first time they open their Grist server: the boot login that proves they have access to the host, and the five-step Quick setup wizard that follows.

The page covers each step in plain language: setting the base URL and edition, picking a sandbox flavor, configuring authentication, hooking up object-storage backups, and the install-wide permission toggles on the apply screen. Screenshots come from a real Grist server captured at 3x DPR in light mode, with each step shown in a representative configuration rather than the bare dev default (Full Grist, gVisor sandbox, getgrist.com auth, MinIO backups).

Also explains how to come back to either page later: the wizard lives in the Admin Panel sidebar permanently, and the boot login is reachable via the "Sign in with boot key" button on the Admin Panel's access-denied card. Linked from mkdocs.yml under the Self-hosted section, right after self-managed.md.


help/en/docs/install/first-run-setup.md
```diff
@@ -0,0 +1,216 @@
...
+**edition** to run.
+
+The base URL matters more than it looks. Auth callbacks, share links, and
+email notifications all build their URLs from it. If you're behind a
+reverse proxy or HTTPS terminator, set it to the public URL users will
+type, not the internal one. Click **Test URL** to have Grist sanity-check
+the value.
+
+The edition toggle picks between the licensed full edition of Grist and the
+free Grist Community edition. See [Self-hosted Grist](../self-managed.md)
```
nbush: This link isn't helpful, but neither is [this](https://support.getgrist.com/self-managed/#how-do-i-enable-the-full-edition-of-grist). We have [this marketing page](https://www.getgrist.com/grist-edition-comparison/) and the [GitHub readme](https://github.com/gristlabs/grist-core#features-in-grist-core) which are better.

nbush (changes requested):
A few higher-level notes, and I can do a tighter review after:
- The bolding for emphasis does look nice throughout, but I'd like to adhere to our existing style guidelines (which uses bolding for column headers). I can do a pass for this once the content is nailed down if you like.
- This line: "Grist asks two questions first: are you really the operator of this server (the boot login page), and how would you like it configured (the Quick setup wizard)?" is followed immediately by the heading "The welcome screen". I like the "two questions" framing, but then I'd like to see those two sections clearly defined after the intro. I think this is as easy as renaming "The welcome screen" to "Boot login page" and having "Set the administrator email" as the last heading _within_ that section.
- Should we link to [this page](https://support.getgrist.com/self-managed/#what-is-the-administrative-account) on the first instance of "administrator"?
- I think we'll need to mention/link this page [here](https://support.getgrist.com/self-managed/#how-do-i-install-grist). If we want to be exhaustive, we'd want to reference the quick setup wizard for all functions mentioned on [this page](https://support.getgrist.com/self-managed/) that are easier to access/change using it.
- I'd also add a small section akin to Admin Controls on this page: https://support.getgrist.com/admin-panel/

Also curious if @nataliemisasi has any other thoughts. But overall, this is very close! Can't wait for this to launch.

help/en/docs/install/first-run-setup.md
```diff
@@ -0,0 +1,216 @@
...
+for the difference.
+
+### Step 2: Sandboxing
+
+Grist runs user-written Python formulas. Without a sandbox, those formulas
+can read and write anything the Grist process can. So on any server that
+hosts documents from more than one author, you want a sandbox on.
```
nataliemisasi: Doesn't read well. Just a few extra words helps 

```suggestion
Grist runs user-written Python formulas. Without a sandbox, those formulas can read and write anything the Grist process can access. Any server hosting documents from more than one author should have the sandbox enabled.
```

help/en/docs/install/first-run-setup.md
```diff
@@ -0,0 +1,216 @@
...
+
+    If `GRIST_SANDBOX_FLAVOR` is already set when Grist starts, the wizard
+    shows the current value and locks the field. To change it, edit your
+    environment and restart.
+
+### Step 3: Authentication
+
+This is the step that matters most for a public-facing server. By default a
+fresh Grist has no authentication. Anyone who reaches it can read and edit
+documents.
```
nataliemisasi:
```suggestion
This is the step that matters most for a public-facing server. By default, a
fresh Grist installation has no authentication. Anyone who reaches it can read and edit
documents.
```

nataliemisasi (commented):
two small updates! This makes me feel like I could attempt to create an installation of my own 😂

paulfitz:
Thanks for the first round of review! Working on incorporating it. One note:

> If we want to be exhaustive, we'd want to reference the quick setup wizard for all functions mentioned on [this page](https://support.getgrist.com/self-managed/) that are easier to access/change using it.

A rule used in the development of the quick setup wizard is that anything configurable from it should also be configurable from the admin page. This rule got broken in one place that I'm patching up now, but after that, there'd never be a reason for someone to need to dig into some random part of the wizard since the thing they need to configure isn't doable from the admin panel.

Separately: admin panel docs need a lot of updates now :)

nataliemisasi (approved):
I made styling updates based on the Grist Content Style Guide. Content looks good to go!

nbush (approved):
Looks good!

PR #653, opened 2026-05-28
Match first-run setup to the wizard's tidied-up labels

paulfitz:
Grist sentence-cased a few buttons and options in the Quick setup wizard, so refresh the screenshots and the one line of text that quoted the old "Go Live" label.

Screenshot updates:

  - setup-server.png: the edition toggle now reads "Community edition" instead of "Community Edition".
  - setup-sandbox.png: the sandbox cards now read "No sandbox" and "macOS sandbox" instead of "No Sandbox" and "macOS Sandbox".
  - setup-apply.png: the step is now titled "Apply & restart" and its button reads "Apply and go live!", both lower-cased from "Apply & Restart" and "Apply and Go Live!".


georgegevoian (approved):
Thanks @paulfitz.

nbush (approved):
👍

PR #658, opened 2026-06-22
Fix some little typos and tidy up a few old links

paulfitz:
A small cleanup pass through the help pages. Fixes a few spelling slips, a couple of links that had stopped working, an example that used slightly the wrong name, and a port number with a typo.

### DIFF
--- a/help/en/docs/mcp.md
+++ b/help/en/docs/mcp.md
@@ -157,7 +157,7 @@ and the underlying scope name is shown in parentheses.
 * **Read documents** (`doc:read`): list and query tables, records, columns, and attachments.
 * **Modify records** (`doc:write`): add, update, and remove rows.
 * **Modify schema** (`doc.schema:write`): add, rename, or remove tables and columns.
-* **Download documents** (`doc:download`): export full documents as CSV or Excel. Not used by any
+* **Download documents** (`doc:download`): download documents in full. Not used by any
   MCP tool currently.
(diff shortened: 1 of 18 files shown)


help/en/docs/self-managed.md
```diff
@@ -681,7 +681,7 @@ client:
   ...
 ```
 
-As per [MinIO specs](https://min.io/docs/minio/linux/developers/go/API.html#:~:text=Default%20value%20is%20us-east,us-east-1).), the default bucket region is `us-east-1`. This default region can be overwritten using the `GRIST_DOCS_MINIO_BUCKET_REGION` flag.
+As per [MinIO specs](https://docs.min.io/community/minio-object-store/developers/go/API.html), the default bucket region is `us-east-1`. This default region can be overwritten using the `GRIST_DOCS_MINIO_BUCKET_REGION` flag.
```
nbush: This redirects to `https://docs.min.io/aistor/developers/`, is that appropriate?

help/en/docs/mcp.md
nbush: We can delete this right?

nbush (changes requested):
A couple tiny changes.

help/en/docs/self-managed.md
paulfitz: minio is such a mess... Linked to something solid in the repo instead of website.

help/en/docs/mcp.md
paulfitz: done

nbush (approved):
👍

PR #667, opened 2026-07-19
Zapier updates

dsagal:
Self-hosted admins on the full edition can give their users one-click, scoped, revocable Zapier connections instead of shared API keys, by standing up a Zapier integration bound to their own instance. Create instructions for them.

The published "Grist" Zapier app now uses OAuth, so the walkthrough's connect step needs to change from "give an API key" to "Connect", and a new screenshot. Include a note pointing self-hosted readers to the "Grist (API key)" app and their own setup options.


help/en/docs/install/integrations/zapier.md
```diff
@@ -0,0 +1,108 @@
+---
+title: Zapier
+---
+
+# Zapier integration {: .tag-ee }
```
paulfitz: So you tag this and explain carefully who it is for and exactly why, but still, given the integrations/zapier path, we should anticipate bots and people of all sorts showing up here wanting to use Grist and Zapier. The top of the page should have a lot more text and images about other easier paths to zapier, so people don't skim and assume they need to clone repos and hack around in many cases. Specially, say something about using Zapier with our SaaS, ideally with an image so they eye doesn't skip over.

It reads like the oauth integration, for small self-hosters, may be a step back in the complexity they need to deal with, as opposed to a step forward?

paulfitz (commented):
Everything makes sense, and perfectly aligns with the technical update. One question about messaging.

help/en/docs/install/integrations/zapier.md
dsagal: Good point about how to start the article. I started wit "Who is this page for?" and included an image of the convenient flow.

The `.tag-ee` isn't a thing any more and shouldn't be there (it's never shown anywhere -- since we made "full grist" the default). I removed.

The oauth integration for small self-hosters intending to use it themselves, is totally a step back :((( It's been a major disappointment. I tried it from all different angles, but Zapier really doesn't offer anything convenient for self-hosters, so I settled on this -- a way for a self-hoster to stand up a convenient integration, which is only really worth it if there are other users who'll use it. It might never get used. (If Zapier had supported DCR or CIMD, the flow could be actually convenient for all.)

paulfitz (approved):
Thanks, good to land once the TODO is done.

PR #681, opened 2026-08-10
Free team sites are limited to 10 team members

paulfitz:
The limits page said there was no limit on team members for team sites on any plan. Free team sites are now capped at 10, so state the cap, what happens at it and over it, and say that guests do not count toward it, since they are capped separately per document.

The August 2022 newsletter announced free team sites with "unlimited team members", and is still served in the archive. Note the limit there too, as a dated note rather than an edit to the announcement, so what was said at the time stays readable.


nataliemisasi (approved):
This looks great. Thanks for adding the note on the old newsletter as well!

PR #689, opened 2026-09-21
Document Grist Fleet

paulfitz:
Add install/fleet.md, covering what Fleet is, how requests reach the server holding a document, and what an installation needs to run one. Link it from the high-availability question in self-managed.md, and add it to the nav.

The page also covers the Servers section of the Admin Panel, with screenshots of the collapsed row, the opened list, a server that has stopped reporting, an installation whose key does not include Fleet, and the settings warning. GET /api/admin/servers gets a section of its own for scripts. Running a fleet, adding a load balancer, and troubleshooting are separated out, so the page reads in the order an operator meets these things rather than as one narrative.


help/en/docs/install/fleet.md
```diff
@@ -0,0 +1,483 @@
...
+### Where users arrive
+
+`APP_HOME_URL` can lead to one of the servers to begin with, with the rest behind
+it. A load balancer in front is recommended, but it does not have to come first:
+add one whenever you are ready and have `APP_HOME_URL` lead there instead, and
+nothing else changes. See [Adding a load balancer](#adding-a-load-balancer).
+
+### Settings to avoid
+
+Some settings are harmless on a single server and quietly wrong across several.
```
nataliemisasi: reword to remove 'quietly'. It's a clear AI-written marker.


```suggestion
Some settings are harmless on a single server but could cause problems across several.
```

nataliemisasi: Not sure if this suggestion I've added is accurate but suggest removing 'quietly' from the sentence.
```suggestion
Some settings are harmless on a single server but could cause problems across several.
```

nataliemisasi (commented):
Overall seems good - makes general sense to a non-technical person like myself.

help/en/docs/install/fleet.md
```diff
@@ -0,0 +1,483 @@
...
+next. Clients using a document on the restarting server reconnect, and the
+document is picked up by a live server and loaded from shared storage.
+
+Without a load balancer, restarting the server your users arrive at interrupts
+everyone until it is back, so leave that one until last. A
+[load balancer](#adding-a-load-balancer) makes the whole upgrade invisible.
+
+### When a server fails
+
+Connections to the documents a failed server held break, and clients reconnect
```
dsagal, line 289: This reads awkwardly. Perhaps: "When a server fails, clients lose connection to any documents held by it, and reconnect automatically..."

help/en/docs/install/fleet.md
```diff
@@ -0,0 +1,483 @@
...
+
+### Documents on one server won't open
+
+If documents will not open while one particular server holds them, even though
+the Admin Panel's **Servers** section shows that server as **running**, the
+address that server published is usually not one the others can use.
+
+Each server works this out for itself, so there is normally nothing to set. Given
+`GRIST_HOST=0.0.0.0`, as in the official Docker images, it listens on every
+interface and advertises the address it reaches Redis on, on the grounds that
```
dsagal, line 402: "advertises the address it reaches Redis on" sounds to me like it's advertising the address of Redis rather than its own. I probably misunderstood, and it may be clearer to those setting up such things. Just mentioning in case there is a simple way to clarify.

dsagal (commented):
I read over this too. This is very useful -- I learned a lot about Fleet! Reads well, just including a couple of comments where it was hard to understand.

PR #690, opened 2026-09-23
Suggest alternatives to MinIO for snapshot storage

paulfitz:
MinIO's open-source edition is no longer maintained, and its images are gone from Docker Hub, but our docs pointed self-hosters at it as the usual store to run next to Grist.

The cloud storage page gains a section on choosing an S3-compatible store. It explains that the store has to support bucket versioning, names Garage and Cloudflare R2 as popular stores that won't work, and lists stores we've tested: AWS S3, RustFS, MinIO, Silo (a MinIO fork) and Versity Gateway. It shows how to create a versioned bucket with the AWS command line tool, which works with any of them.

The snapshots walkthrough in the self-managed page now runs RustFS and creates the versioned bucket with the amazon/aws-cli image, instead of sending you to MinIO's web console. The first-run setup page names RustFS instead of MinIO as an example store.

The docs use the GRIST_DOCS_S3_* setting names, which Grist accepts alongside the older GRIST_DOCS_MINIO_* ones. The old names are mentioned once, since older versions of Grist need them.

The cloud storage page also said Grist prints a line at startup showing which store is active. It only does that with DEBUG=1 now.


paulfitz:
(should not land before a companion code change has landed, making some of the changes documented)

